PR #2, opened 2021-03-11
iOS Swift Interface Builder template starter app client

cbush:
## Template Starter App Client Design Principles

- *Minimal over nice:* The intended audience of this template will build their app on top of this app. Any "nice" thing that adds complexity (whether additional lines of code, files, functions, etc.) will probably have to be peeled out by the user. Therefore, "nice" runs counter to the goal of these templates. "Nice" should go in our app gallery.
- *Correct over minimal:* These apps might become the basis for many future production apps. Do not cut corners for the sake of simplicity. Do things properly according to the idioms, guidelines, and best practices of the target platform.

## Screenshots

Storyboard (simple):
image

Welcome screen:
image

Tasks screen (updated with label):
image

Tasks screen (add task):
image

Tasks list (check, swipe to delete):
image


MongoCaleb (approved):
LGTM. Do you intend to keep the appID in there?

cbush:
@MongoCaleb I did. But it's a good question. Should we commit this here? It will ultimately get replaced before it reaches the end user upon deployment.

ios-swift/App/WelcomeViewController.swift
```diff
@@ -0,0 +1,136 @@
...
+        }
+    }
+    
+    /// Called when sign in completes. Opens the realm asynchronously and navigates to the Tasks screen.
+    func onSignInComplete(_ user: User) {
+        setDefaultConfiguration(forUser: user)
+
+        setLoading(true)
+        Realm.asyncOpen { result in
+            DispatchQueue.main.async {
```
tgoyne, line 94: `asyncOpen()` calls the callback on the main queue if no queue is specified.

RashmiSrinivas19:
@cbush This looks good to me pending one item missing - 

In the designs, we have some explanatory text at the [bottom of the screen](https://www.figma.com/file/mjE5GFT1iU9g3Drd5fl5at/Copy-of-Template-Apps-UI?node-id=19%3A5732) - is it possible to get that into the task view?

cbush:
@RashmiSrinivas19 ah yes, I can add that!

RashmiSrinivas19:
@cbush Looks great to me!

cbush:
@RashmiSrinivas19 suggested adding a shadow to indicate whether content can be found below the fold. Doesn't add much code so I think it's great.
image

PR #5, opened 2021-03-22
DOCSP-14516 Android - Implement welcome screen

jenniferpeshansky:
Generated a blank Android app using Kotlin and Gradle. I probably still need to strip stuff out from this eventually.


.gitignore
```diff
@@ -1,2 +1,10 @@
+*.iml
+.gradle
+/local.properties
+/.idea
 .DS_Store
+/build
+/example/build
+/annotation-processor/build
+/captures
```
cbush, line 9: Are these changes needed? Any of these things should be in the android-kotlin/.gitignore specifically.

jenniferpeshansky: oh, didn't realize I had changed the main .gitignore file, I'll revert

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
```diff
@@ -0,0 +1,125 @@
...
+
+    override fun onBackPressed() {
+        // Disable going back to the MainActivity
+        moveTaskToBack(true)
+    }
+
+    private fun onLoginSuccess() {
+        // successful login ends this activity, bringing the user back to the project activity
+        startActivity(Intent(this, ItemsActivity::class.java))
+        // finish()
```
cbush, line 45: Is this needed?

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
```diff
@@ -0,0 +1,125 @@
...
+        Log.e(TAG(), errorMsg)
+        Toast.makeText(baseContext, errorMsg, Toast.LENGTH_LONG).show()
+    }
+
+    // Function to validate that the username adheres to all username rules.
+    private fun validateUsername(username: String): Boolean {
+        if (username.isEmpty()) {
+            return false // An empty username is not valid.
+        }
+        return true
```
cbush, line 58: How about...

```
return !username.isEmpty()
```

Or:

```
// Add additional checks here.
return true
```

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    private fun onLoginFailed(errorMsg: String) {
+        Log.e(TAG(), errorMsg)
+        Toast.makeText(baseContext, errorMsg, Toast.LENGTH_LONG).show()
+    }
+
+    // Function to validate that the username adheres to all username rules.
+    private fun validateUsername(username: String): Boolean {
+        if (username.isEmpty()) {
```
cbush, line 55: I don't know if we want to include username validation at all. The server will do some validation.

By the way, you might want to use `isBlank()` here.

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
```diff
@@ -0,0 +1,125 @@
...
+            return false // An empty username is not valid.
+        }
+        return true
+    }
+
+    // Function to validate that the password adheres to all password rules.
+    private fun validatePassword(password: String): Boolean {
```
cbush, line 62: Same, I don't know if we want to add any validation at all. Users can add their own validation later as they build up the app and the server will do some minimal validation.

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
```diff
@@ -0,0 +1,125 @@
...
+        if (password.isEmpty()) {
+            return false // An empty password is not valid.
+        }
+        return true
+    }
+
+    // Function to swap the log in use case with the create account use case.
+    private fun toggleExistingUser() {
```
cbush, line 70: The name of this method could stand to match the name of the variable it is updating so that we have a better idea what it does. As I mentioned in another comment, I would mention "mode" or something like that so that it's clear this has to do with the mode the UI is in rather than toggling an actual user.

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
```diff
@@ -0,0 +1,125 @@
...
+
+    // Function to validate that the password adheres to all password rules.
+    private fun validatePassword(password: String): Boolean {
+        if (password.isEmpty()) {
+            return false // An empty password is not valid.
+        }
+        return true
+    }
+
+    // Function to swap the log in use case with the create account use case.
```
cbush, line 69: It might be better to use javadoc notation (`/** ... */`) here. Not that people will generate javadoc on this.

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
```diff
@@ -0,0 +1,125 @@
...
+    // handle user authentication (login) and account creation
+    private fun login() {
+        // while this operation completes, disable the button to login or create a new account
+        loginButton.isEnabled = false
+
+        val username = this.username.text.toString()
+        val password = this.password.text.toString()
+
+        if (!validateUsername(username)) {
+            onLoginFailed("Invalid username")
```
cbush, line 90: You forgot to reenable the loginButton.

It's best to be very careful when handling state like this. I would move the UI change to the very last moment before actually going async.

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
```diff
@@ -0,0 +1,125 @@
...
+            // register a user using the Realm App
+            myApp.emailPassword.registerUserAsync(username, password) {
+                // re-enable the buttons after user registration returns a result
+                loginButton.isEnabled = true
+                if (!it.isSuccess) {
+                    onLoginFailed( "Error: ${it.error}")
+                } else {
+                    Log.i(TAG(), "Successfully registered user.")
+                    // when the account has been created successfully, log in to the account
+                    this.createUser = false
```
cbush, line 108: Changing the mode and calling "login" again doesn't feel very direct and intentional. Ideally, you could rename this `login()` method to something like `onLoginButtonClicked`, which is more descriptive of what it does. Then put `logIn()` and `registerUser()` in their own methods, so that you can call `logIn()` directly from the success handler of `registerUser()`. This would clean up the code by reducing the indentation level and also make it more intentional.

cbush (changes requested):
Looks great so far. I have several mostly style comments. Enough to warrant requesting changes as I'd like to take another look before merging.

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
```diff
@@ -0,0 +1,125 @@
...
+import android.content.Intent
+import androidx.appcompat.app.AppCompatActivity
+import android.os.Bundle
+import android.util.Log
+import android.widget.Button
+import android.widget.EditText
+import android.widget.Toast
+import io.realm.mongodb.Credentials
+/*
+* LoginActivity: launched whenever a user isn't already logged in. Allows a user to enter email
```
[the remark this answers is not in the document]
jenniferpeshansky: fixed!

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
```diff
@@ -0,0 +1,125 @@
...
+* and password credentials to log in to an existing account or create a new account.
+*/
+class LoginActivity : AppCompatActivity() {
+    private lateinit var username: EditText
+    private lateinit var password: EditText
+    private lateinit var loginButton: Button
+    private lateinit var accountExistsToggleButton: Button
+    private var createUser = true
```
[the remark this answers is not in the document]
jenniferpeshansky: I like isCreateUserMode, changed it

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
jenniferpeshansky: I was still deciding whether or not to have the items as the main screen and open the login screen over it (as is done in the tutorial app), or to open the login screen first and open the items screen from there. I'll choose one and fix the comment to reflect that, but I'm curious to hear your thoughts on which is simpler

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
jenniferpeshansky: Even in the tutorial app, wouldn't it be dangerous to allow an empty username and password? That is, will it crash the backend?

jenniferpeshansky: I tested it out and it looks like the server does complain (rather dramatically) when the username and password are blank, so I guess we don't need to validate at all unless we want the empty functions

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
jenniferpeshansky: that makes sense, updating it to match

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
jenniferpeshansky: it looks prettier in javadoc notation, at any rate

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
jenniferpeshansky: makes sense, I did so

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
cbush: Whichever you prefer is fine, but I would delete this line rather than leaving it commented.

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
cbush: I don't think client-side validation is needed here, but I don't recall what we do on the iOS app.

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
jenniferpeshansky: I agree, I've cleaned up a lot of the functions to be more clear

android-kotlin/app/src/main/java/com/mongodb/app/ItemsActivity.kt
```diff
@@ -0,0 +1,37 @@
...
+import android.content.Intent
+import androidx.appcompat.app.AppCompatActivity
+import android.os.Bundle
+import android.util.Log
+import android.widget.Button
+import io.realm.Realm
+
+class ItemsActivity : AppCompatActivity() {
+    private lateinit var logoutButton: Button
+    private var user: io.realm.mongodb.User? = null
```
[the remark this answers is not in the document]
jenniferpeshansky: I refactored to use myApp.currentUser() for now. What do you mean by computed property?

android-kotlin/app/src/main/java/com/mongodb/app/ItemsActivity.kt
```diff
@@ -0,0 +1,37 @@
...
+        if (user == null) {
+            startActivity(Intent(this, LoginActivity::class.java))
+        }
+        logoutButton = findViewById(R.id.button_log_out)
+        logoutButton.isEnabled = true
+        logoutButton.setOnClickListener{ (logout()) }
+    }
+
+    private fun logout() {
+        user?.logOutAsync {
```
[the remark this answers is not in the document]
jenniferpeshansky: I'm not sure what the purpose of disabling and re-enabling is besides to avoid multiple calls that all generate errors if the user spams the logout button fast enough. Since this isn't likely, is the disabling and re-enabling unnecessary? or is it just good practice?

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
jenniferpeshansky: removed all validation

android-kotlin/app/src/main/java/com/mongodb/app/MyApp.kt
```diff
@@ -0,0 +1,39 @@
...
+import io.realm.mongodb.App
+import io.realm.mongodb.AppConfiguration
+
+lateinit var myApp: App
+
+// global Kotlin extension that resolves to the short version
+// of the name of the current class. Used for labelling logs.
+inline fun <reified T> T.TAG(): String = T::class.java.simpleName
+const val PARTITION_EXTRA_KEY = "PARTITION"
+const val PROJECT_NAME_EXTRA_KEY = "PROJECT NAME"
```
cbush, line 18: You can probably delete both of these lines.

android-kotlin/app/src/main/java/com/mongodb/app/MyApp.kt
```diff
@@ -0,0 +1,39 @@
...
+import android.app.Application
+import android.util.Log
+
+import io.realm.Realm
+import io.realm.log.LogLevel
+import io.realm.log.RealmLog
+import io.realm.mongodb.App
+import io.realm.mongodb.AppConfiguration
+
+lateinit var myApp: App
```
cbush, line 12: Minor: I would actually rename this to `realmApp`. `my`... is great for tutorials, but in a real app I wouldn't expect to see it so much.

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
```diff
@@ -0,0 +1,108 @@
...
+            loginButton.setText(getString(R.string.create_account))
+            accountExistsToggleButton.setText(getString(R.string.already_have_account))
+        } else {
+            loginButton.setText(getString(R.string.log_in))
+            accountExistsToggleButton.setText(getString(R.string.does_not_have_account))
+        }
+    }
+
+    /**
+     *  Log an error and propagate it to the user
```
cbush, line 102: Minor, but it's best to prefer simpler language:
```suggestion
     *  Log an error and present it to the user
```

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
```diff
@@ -0,0 +1,108 @@
...
+            }
+        }
+    }
+
+    /**
+     *  Authenticate a user
+     */
+    private fun login(username: String, password: String) {
+        val creds = Credentials.emailPassword(username, password)
+        myApp.loginAsync(creds) {
```
cbush, line 75: Did you disable loginButton in this method?

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     *  Handle account creation and user authentication
+     */
+    private fun onLoginButtonClicked() {
+        val username = this.username.text.toString()
+        val password = this.password.text.toString()
+        if (isCreateUserMode) {
+            registerUser(username, password)
+        } else {
+           login(username, password)
```
cbush, line 46: Is this indented correctly?
```suggestion
            login(username, password)
```

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
```diff
@@ -0,0 +1,108 @@
...
+        username = findViewById(R.id.input_username)
+        password = findViewById(R.id.input_password)
+
+        loginButton = findViewById(R.id.button_create_login)
+        loginButton.setOnClickListener { onLoginButtonClicked() }
+        loginButton.isEnabled = true
+
+        accountExistsToggleButton = findViewById(R.id.button_existing_account)
+        accountExistsToggleButton.setOnClickListener { toggleCreateUserMode() }
+        accountExistsToggleButton.isEnabled = true
```
cbush, line 34: Nice cleanup!

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
cbush: Nitpick: `logIn()` would be correct. "Log in" is the verb, "login" is the noun.

cbush (approved):
Looks good. Just a few additional comments to check. Thanks!

android-kotlin/app/src/main/java/com/mongodb/app/MyApp.kt
jenniferpeshansky: done

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
jenniferpeshansky: fixed

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
jenniferpeshansky: fixed

android-kotlin/app/src/main/java/com/mongodb/app/LoginActivity.kt
jenniferpeshansky: :)

cbush (approved):
LGTM! Feel free to merge.

PR #7, opened 2021-03-30
DOCSP-14517 Android - Implement items screen

jenniferpeshansky:
WIP for now since the task bar functionality is limited, but the items screen is working.


android-kotlin/app/src/main/java/com/mongodb/app/ItemAdapter.kt
```diff
@@ -0,0 +1,33 @@
...
+import androidx.recyclerview.widget.RecyclerView
+import io.realm.OrderedRealmCollection
+import io.realm.Realm
+import io.realm.RealmRecyclerViewAdapter
+import io.realm.kotlin.where
+import io.realm.mongodb.sync.SyncConfiguration
+import org.bson.types.ObjectId
+
+/**
+ * ItemAdapter: extends the Realm-provided RealmRecyclerViewAdapter to provide
```
cbush, line 16: No need to introduce in the comment.
```suggestion
 * Extends the Realm-provided RealmRecyclerViewAdapter to provide
```

android-kotlin/app/src/main/java/com/mongodb/app/ItemsActivity.kt
```diff
@@ -1,39 +1,154 @@
...
+    override fun onStart() {
+        super.onStart()
+        val user = realmApp.currentUser()
+        if (user == null) {
+            // if no user is currently logged in, start the login activity so the user can authenticate
+            startActivity(Intent(this, LoginActivity::class.java))
+        }
+        else {
+            // get the partition value and name of the project we are currently viewing
+            val partition = realmApp.currentUser().toString()
```
cbush, line 39: I'm not 100% familiar with the Java/Kotlin SDK so I'm not sure, but this doesn't quite look correct. I think the partition value should be the current user's ID. This looks like it might serialize the actual user object. (Which would technically "work" until you try to see your tasks on the iOS client, which does use the user ID). Can you confirm?

android-kotlin/app/src/main/java/com/mongodb/app/ItemsActivity.kt
```diff
@@ -1,39 +1,154 @@
...
+    // Ensure the user realm closes when the activity ends.
+    override fun onDestroy() {
+        super.onDestroy()
+        userRealm?.close()
+        recyclerView.adapter = null
+    }
+    override fun onStop() {
+        super.onStop()
+        realmApp.currentUser().run {
+            userRealm?.close()
```
cbush, line 82: Seems odd that the realm is closed in both the destroy and stop lifecycle methods. Shouldn't we pick one?

android-kotlin/app/src/main/java/com/mongodb/app/ItemsActivity.kt
```diff
@@ -1,39 +1,154 @@
...
+    override fun onCreateOptionsMenu(menu: Menu): Boolean {
+        menuInflater.inflate(R.menu.task_menu, menu)
+        return true
+    }
+
+    /**
+     *  Logs out the user and brings them back to the login screen.
+     */
+    override fun onOptionsItemSelected(item: MenuItem): Boolean {
+        val user = realmApp.currentUser()
```
cbush, line 95: This is declared well before needed.

android-kotlin/app/src/main/java/com/mongodb/app/ItemsActivity.kt
```diff
@@ -1,39 +1,154 @@
...
+        return when (item.itemId) {
+            R.id.action_logout -> {
+                user?.logOutAsync {
+                    if (it.isSuccess) {
+                        realmApp.removeUser(realmApp.currentUser())
```
cbush, line 100: Why do you remove the user? This doesn't look like it will work as expected. After log out, the current user should be null. So, I'm not sure what the intention is here.

android-kotlin/app/src/main/java/com/mongodb/app/ItemsActivity.kt
```diff
@@ -1,39 +1,154 @@
...
+                    dialog.cancel()
+                }
+        val dialog = dialogBuilder.create()
+        dialog.setView(input)
+        dialog.setTitle("Add Item")
+        dialog.show()
+        input.requestFocus()
+    }
+
+    private fun setUpRecyclerView(realm: Realm, user: User?, partition: String) {
```
cbush, line 145: This API feels unintentional.

- We have the realm is the userRealm member
- We only call this when we know the user is logged in, so the user doesn't have to be optional. Besides, we have global access to the realmApp and can always get the currentUser() from that.
- The partition should be readable from the realm itself (not sure on Kotlin, but it is in Swift). Besides, the partition value doesn't seem to be used by the adapter?

I would recommend rethinking this method a bit.

cbush (changes requested):
Great start! A few questions.

android-kotlin/app/src/main/java/com/mongodb/app/ItemsActivity.kt
jenniferpeshansky: yep, using Destroy

android-kotlin/app/src/main/java/com/mongodb/app/ItemsActivity.kt
jenniferpeshansky: you're right, I removed it from being a local variable in this method at all

android-kotlin/app/src/main/java/com/mongodb/app/ItemsActivity.kt
jenniferpeshansky: true - I was trying to adapt it but I think I misunderstood what the goal was

android-kotlin/app/src/main/java/com/mongodb/app/ItemsActivity.kt
jenniferpeshansky: I decided to inline this functionality into the onStart method, let me know your thoughts.

cbush (approved):
Excellent!

PR #8, opened 2021-03-31
Web Template App (GraphQL, MQL, Local Mock)

nlarew:
## TODO

- [x] Run `npm run generate:prod` and commit the output artifact projects
- [x] Update README
- [x] Add text/links to the bottom of each page
- [x] Source API type (`REACT_APP_API_TYPE`) from a `.env` file
- [x] Contextualize app title to the API type
- [x] Determine the href for "GraphiQL console" 
   - What's the best way to get the project/app ids? 
   - https://realm.mongodb.com/groups/604663cdf7eac57d66db8a1c/apps/604664b7083e1671cfd981d4/graphql/explore
- [x] Update theme colors


web-js/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+```shell
+gh repo clone mongodb-university/realm-template-apps
+cd realm-template-apps/web-js
+npm install
+```
+
+### 2. Define `.env` Values
+
+Open the `.env` file and define the following values:
```
mongomoe, line 26:
```suggestion
Open the `.env` file and uncomment one of the following values:
```

web-js/.env
```diff
@@ -0,0 +1,4 @@
+REACT_APP_API_TYPE="graphql"
+# REACT_APP_API_TYPE="local"
+# REACT_APP_API_TYPE="mql"
+REACT_APP_REALM_APP_ID="todo-sync-tmikv"
```
mongomoe, line 4: App id should be stored in a 'realm.json' file per the spec.

[Spec: Template Starter Applications > "Each framework will require its own format"](https://docs.google.com/document/d/1gmbjUjeSULW4aI3OV9ytBqU0sypQ78H2cpIKIHr1xWM/edit?usp=sharing)

Screen Shot 2021-04-23 at 3 29 01 PM


You can copy the one in the react-native template - https://github.com/mongodb-university/realm-template-apps/blob/react-native/react-native/realm.json

web-js/src/components/WelcomePage.js
```diff
@@ -0,0 +1,155 @@
...
+  const clearErrors = () => setError(noErrors);
+  // Manage password visibility
+  const [showPassword, setShowPassword] = React.useState(false);
+  const toggleShowPassword = () => setShowPassword((s) => !s);
+
+  const onFormSubmit = async ({ email, password }) => {
+    clearErrors();
+    try {
+      if (isSignup) {
+        await realmApp.emailPasswordAuth.registerUser(email, password);
```
mongomoe, line 41: When a user creates an account, they don't see anything in the UI that the account has been successfully created. This lead me to click the button twice and get an error saying the account has already been created. 

Can you do one of the following:
* Add some UI message saying 'your account has been successfully created' once it has been created.
or 
* Log the user in after the user has registered

web-js/src/components/WelcomePage.js
```diff
@@ -0,0 +1,155 @@
...
+              : "Sign up for an account"}
+          </button>
+        </form>
+      </Card>
+      <MoreInfoTemplateAndDocs />
+    </Container>
+  );
+}
+
+function parseAuthenticationError(err) {
```
mongomoe, line 119: Can we either extract these functions into a separate file so they can be used in other files as helper methods or have them as a method inside of the WelcomePage functional component (similar to the methods clearErrors, and onFormSubmit) in the case that these are only to be used by this component?

web-js/src/components/WelcomePage.js
```diff
@@ -0,0 +1,155 @@
...
+  const parts = err.message.split(":");
+  const reason = parts[parts.length - 1].trimStart();
+  if (!reason) return { status: "", message: "" };
+  const reasonRegex = /(?<message>.+)\s\(status (?<status>[0-9][0-9][0-9])/;
```
mongomoe, line 123: Dang it sucks that we have to regex parse these errors, instead of there being error codes.

web-js/src/components/TodoItemsPage.js
```diff
@@ -0,0 +1,139 @@
...
+          </List>
+        </div>
+      )}
+      <MoreInfoGraphiQL />
+      <MoreInfoTemplateAndDocs />
+    </Container>
+  );
+}
+
+function TodoItem({ isDraft = false, todo, todoActions }) {
```
mongomoe, line 70: Did you mean to do something with the `isDraft` variable? It's not used anywhere here.

web-js/src/components/util-hooks.js
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+
+export function useTimeout(fn, ms) {
+  React.useEffect(() => {
+    const timeout = setTimeout(() => { fn() }, ms)
+    return () => clearTimeout(timeout)
+  }, [fn, ms])
+}
+
+export function useShowLoader(loading, delayMs) {
```
mongomoe, line 10: This looks beautiful in the UI!

web-js/src/components/TodoItemsPage.js
```diff
@@ -0,0 +1,139 @@
...
+import { useShowLoader } from "./util-hooks";
+import { MoreInfoTemplateAndDocs, MoreInfoGraphiQL } from './MoreInfo'
+
+export function TodoItemsPage() {
+  const { loading, todos, ...todoActions } = useTodos();
+  const { draftTodos, ...draftTodoActions } = useDraftTodos();
+  const showLoader = useShowLoader(loading, 200);
+  return (
+    <Container className="main-container" maxWidth="sm">
+      {loading ? (showLoader ? <LinearProgress /> : null) : (
```
mongomoe, line 30: This looks beautiful in the UI!

web-js/src/components/TodoItemsPage.js
```diff
@@ -0,0 +1,139 @@
...
+    </ListItem>
+  );
+}
+
+const ListItemWithTwoSecondaryActions = withStyles({
+  secondaryAction: {
+    paddingRight: "120px",
+  },
+})(ListItem);
+function DraftTodoItem({ todo, todoActions, draftTodoActions }) {
```
mongomoe, line 101:
```suggestion

function DraftTodoItem({ todo, todoActions, draftTodoActions }) {
```

web-js/src/components/TodoItemsPage.js
mongomoe, line 70: Can this and the DraftTodoItem component be moved to their own files? Seems weird to have them in this file.  Alternatively, if you suspect these components aren't going to be reused you could have them as methods of the 'TodoItemsPage' component, this would better indicate that they're only to be used in this component.

mongomoe:
Question: Why does `src/components` contain things that aren't react components?

web-js/src/components/WelcomePage.js
nlarew: https://jira.mongodb.org/browse/RJS-981

^ Totally forgot but they actually do implement it for realm-web already. I'll update to use it :)

web-js/src/components/WelcomePage.js
nlarew: Why? They're only used in this file + don't rely on props so there's no need to include them in the function closure.

I'd rather have a smaller component function and leave this at the bottom of the file for the adventurous to explore.

web-js/src/components/WelcomePage.js
nlarew: Good catch! Updated to auto-login on successful signup.

web-js/src/components/TodoItemsPage.js
nlarew: Good catch - missed this in a refactor

nlarew: Moved to separate files

web-js/src/components/RealmApp.js
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+import * as Realm from "realm-web";
+
+const RealmAppContext = React.createContext(null);
+
+export function RealmAppProvider({ appId, children }) {
```
mongomoe, line 6: Does 'RealmAppProvider' need to exist for this application? It seems it's functionality is basically: 

* logIn
* logOut
* pass logged in user to other components
* pass the Realm App to other components. 

An alternative approach to a 'RealmAppProvider' component is to do the following:
* The actual logIn/logOut could be methods on a component (probably WelcomePage and App for this application)
* You can get the logged-in user by calling '[Realm.App.currentUser]'(https://docs.mongodb.com/realm-sdks/js/latest/Realm.App.html#currentUser) anywhere in your application. 
* You can get the Realm App anywhere in your application by calling '[Realm.App.getApp](https://docs.mongodb.com/realm-sdks/js/latest/Realm.App.html#getApp)' 


This alternative approach offers simplicity (which is a major goal of this project since it is a template app and users may be confused by any additional/unnecessary bulk), and parity with the react native template app).

You can see the implementation for React Native here: https://github.com/mongodb-university/realm-template-apps/blob/main/react-native/src (let me know if you have questions)

web-js/src/components/useDraftTodos.js
```diff
@@ -0,0 +1,40 @@
...
+  const createDraftTodo = () => {
+    const draftTodo = {
+      _id: new Realm.BSON.ObjectID(),
+      summary: "",
+      isComplete: false,
+    };
+    setDrafts((d) => [...d, draftTodo]);
+  };
+
+  const setDraftTodoSummary = (draft, summary) => {
```
mongomoe, line 16: can you add some comments to this method (and actually probably all the methods in this file). Especially lines like: 
```
      return [
        ...oldDrafts.slice(0, idx),
        { ...oldDrafts[idx], summary },
        ...oldDrafts.slice(idx + 1),
      ];
```

and 

```
      const idx = oldDrafts.findIndex((d) => d._id === draft._id);
      return [...oldDrafts.slice(0, idx), ...oldDrafts.slice(idx + 1)];
```
These lines are hard to parse/scan quickly and **need** comments,

web-js/src/components/useTodos.js
```diff
@@ -0,0 +1,26 @@
...
+  case "mql": {
+    useTodos = useMqlTodos
+    break;
+  }
+  case "local": {
+    useTodos = useLocalTodos
+    break;
+  }
+  default: {
+    throw new Error(`Invalid REACT_APP_API_TYPE: "${API_TYPE}". Specifiy "graphql", "mql", or "local" instead.`)
```
mongomoe, line 22:
```suggestion
    throw new Error(`Invalid REACT_APP_API_TYPE: "${API_TYPE}". Specify "graphql", "mql", or "local" instead.`)
```

web-js/src/components/useTodos_graphql.js
```diff
@@ -0,0 +1,164 @@
...
+              isComplete
+              summary
+            }
+          }
+        `,
+      });
+      return data.tasks
+    };
+
+    const watchTodos = async () => {
```
mongomoe, line 68: Can you add more comments to this method? Explaining what the loops are doing, what the anonymous functions are doing, and what is being returned.

web-js/src/components/useTodos_local.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export function useTodos() {
+  const realmApp = useRealmApp();
+  const [todos, setTodos] = React.useState([]);
+  
+  const [loading, setLoading] = React.useState(true);
+  React.useEffect(() => {
+    const fetchTodos = async () => {
+      await latency(1640)
```
mongomoe, line 31: What's up with this artificial timer?

web-js/src/components/useTodos_mql.js
```diff
@@ -0,0 +1,81 @@
...
+  }, [realmApp.currentUser]);
+  const [todos, setTodos] = React.useState([]);
+  
+  const [loading, setLoading] = React.useState(true);
+  React.useEffect(() => {
+    const fetchTodos = async () => {
+      return await taskCollection.find({});
+    }
+    
+    const watchTodos = async () => {
```
mongomoe, line 18: can you add some comments here

web-js/src/components/useTodos_graphql.js
```diff
@@ -0,0 +1,164 @@
...
+          }
+          case "delete": {
+            setTodos((oldTodos) => {
+              const idx = getTodoIndex(oldTodos);
+              return [...oldTodos.slice(0, idx), ...oldTodos.slice(idx + 1)];
+            });
+            break;
+          }
+          default: {
+            // change.operationType will always be one of the specified cases, so we should never hit this default
```
mongomoe, line 99: If you should never hit this default, perhaps you should just remove this blank default or have it throw an error?

web-js/src/components/useTodos_graphql.js
mongomoe, line 68: Is `watchTodos` the same for the files 'useTodos_graphql.js' and 'useTodos_mql.js'? 

If so, you should abstract these methods somehow to make them more [D.R.Y](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

web-js/src/components/useTodos_graphql.js
```diff
@@ -0,0 +1,164 @@
...
+    });
+  }, [realmApp.currentUser]);
+
+  return client;
+}
+
+export function useTodos() {
+  const realmApp = useRealmApp();
+  const graphql = useApolloClient();
+  const taskCollection = React.useMemo(() => {
```
mongomoe, line 44: The code from lines 44 - 50 (from defining taskCollection to creating state for loading) seems to be the same as the code in 'useTodos_mql' (lines 6-12). Can this code be abstracted to be more [D.R.Y](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)?

web-js/src/components/useTodos_local.js
```diff
@@ -0,0 +1,67 @@
...
+    fetchTodos().then((t) => {
+      setTodos(t);
+      setLoading(false);
+    });
+  }, [realmApp.currentUser.id]);
+  
+  const saveTodo = async (todo) => {
+    setTodos((t) => [...t, todo]);
+  };
+  const toggleTodo = async (todo) => {
```
mongomoe, line 43: can you add comments to any methods in this file that are doing slicing, and returning arrays with logic?

web-js/src/components/TodoItemsPage.js
```diff
@@ -0,0 +1,61 @@
...
+import { DraftTodoItem } from './DraftTodoItem'
+import { useTodos } from "./useTodos";
+import { useDraftTodos } from "./useDraftTodos";
+import { useShowLoader } from "./util-hooks";
+import { MoreInfoTemplateAndDocs, MoreInfoGraphiQL } from './MoreInfo'
+
+export function TodoItemsPage() {
+  const { loading, todos, ...todoActions } = useTodos();
+  const { draftTodos, ...draftTodoActions } = useDraftTodos();
+  const showLoader = useShowLoader(loading, 200);
```
mongomoe, line 20: I'm a bit confused by this loader. Why the artificial 200ms loader?

mongomoe (commented):
Can you add comments to any method performing a Grapqhl statement? Maybe just a quick one-line explaining what the method does, unless it's super self-explanatory

web-js/src/components/WelcomePage.js
```diff
@@ -0,0 +1,144 @@
...
+import Visibility from "@material-ui/icons/Visibility";
+import VisibilityOff from "@material-ui/icons/VisibilityOff";
+
+export function WelcomePage() {
+  const realmApp = useRealmApp();
+  // Track whether the user is logging in or signing up for a new account
+  const [isSignup, setIsSignup] = React.useState(false);
+  const toggleIsSignup = () => {
+    clearErrors();
+    setIsSignup((s) => !s);
```
mongomoe, line 24: Let's change the variable to be a bit more clear what 's' is here.

web-js/src/components/WelcomePage.js
```diff
@@ -0,0 +1,144 @@
...
+  // Authentication errors
+  const noErrors = {
+    email: null,
+    password: null,
+  };
+  const [error, setError] = React.useState(noErrors);
+  const clearErrors = () => setError(noErrors);
+  // Manage password visibility
+  const [showPassword, setShowPassword] = React.useState(false);
+  const toggleShowPassword = () => setShowPassword((s) => !s);
```
mongomoe, line 35: Let's change the variable to be a bit more clear what 's' is here.

mongomoe (commented):
UI looks beautiful, awesome job on this. 

Most of my suggestions are to: 
- Add more comments to methods that aren't self-explanatory
- Make the code more D.R.Y by abstracting code that is similar/the same to helper methods 

Also:
- Replace bulky code such as 'RealmAppProvider' with SDK methods 

Let me know if you have any questions

web-js/src/components/RealmApp.js
nlarew: Does it strictly _need_ to exist? No. But it does include a few advantages when you view the template as a base app devs can build from:

- **Singleton definition & instantiation** - the RN template instantiates the App several times (once per view)
- **Composability & reuse** - some folks use multiple Realm apps and this approach lets you dynamically define multiple instead of hardcoding one app at the module level.
- **State encapsulation** - the RN template mirrors the `user` as state in `WelcomeView` and would need to do this for any component where auth changes should affect app state (i.e. really any auth change). Along the lines of the singleton point, this encapsulates the state update in a single spot.
- **Future-poofing** - our tentative plan is to use the provider/consumer approach in the official Realm hooks, so this API should hopefully not require significant changes to adapt the official package whenever we release it.

web-js/src/components/useTodos_local.js
nlarew: This is the "local" todo provider which mocks everything in-memory and doesn't actually connect to the database. The timer is helpful for testing network latency - in this specific case I think I was testing the loading bar.

web-js/src/components/useTodos_graphql.js
nlarew: Ah yeah it should throw an error - good catch

web-js/src/components/TodoItemsPage.js
nlarew: The 200ms here means that the loader should only show if the request takes longer than 200ms to complete. It's a similar idea to React Suspense - it's jarring and unnecessary to see loaders for things that load fast, so you should wait a little bit to render them.

web-js/src/components/RealmApp.js
```diff
@@ -0,0 +1,46 @@
...
+  const logIn = React.useCallback(
+    async (credentials) => {
+      await realmApp.logIn(credentials);
+      setCurrentUser(realmApp.currentUser);
+    },
+    [realmApp]
+  );
+  // Wrap the current user's logOut function to remove the logged out user from state
+  const logOut = React.useCallback(async () => {
+    await realmApp.currentUser?.logOut();
```
kneth, line 24: Should we remove the user by calling `Realm.App.removeUser()`?

web-js/src/components/WelcomePage.js
```diff
@@ -0,0 +1,146 @@
...
+  IconButton,
+  Card,
+  Typography,
+  InputAdornment,
+} from "@material-ui/core";
+import { useRealmApp } from "./RealmApp";
+import { MoreInfoTemplateAndDocs } from './MoreInfo'
+import Visibility from "@material-ui/icons/Visibility";
+import VisibilityOff from "@material-ui/icons/VisibilityOff";
+import { toggleBoolean } from "../utils";
```
kraenhansen, line 16: I suggest separating imports of external packages vs relative files into two sections (external first) to improve readability here.

web-js/src/components/WelcomePage.js
```diff
@@ -0,0 +1,146 @@
...
+      <Card className="auth-card" variant="outlined">
+        <form
+          className="auth-form"
+          onSubmit={(e) => {
+            e.preventDefault();
+            function getFormValues(htmlFormElement) {
+              const formData = new FormData(htmlFormElement);
+              return Object.fromEntries(formData.entries());
+            }
+            const { email, password } = getFormValues(e.target);
```
kraenhansen, line 62: Seems a little excessive to create an inline function for this.
```suggestion
            const formData = new FormData(e.target);
            const { email, password } = Object.fromEntries(formData.entries());
```

web-js/src/components/WelcomePage.js
```diff
@@ -0,0 +1,146 @@
...
+              : "Sign up for an account"}
+          </button>
+        </form>
+      </Card>
+      <MoreInfoTemplateAndDocs />
+    </Container>
+  );
+}
+
+function handleAuthenticationError(err, setError) {
```
kraenhansen, line 120: What happens if the `err` is not a `MongoDBRealmError` (like a network level error or something)? I believe the error would be swallowed, since the `err.error` and `err.statusCode` will both be `undefined` and the `default` case doesn't call `setError`.

web-js/src/hooks/useTodos.js
```diff
@@ -0,0 +1,31 @@
...
+    useTodos = useLocalTodos
+    break;
+  }
+  default: {
+    throw new Error(`Invalid REACT_APP_API_TYPE: "${API_TYPE}". Specify "graphql", "mql", or "local" instead.`)
+  }
+}
+// :state-end:
+// :state-uncomment-start: prod-mql
+import { useTodos } from "./useTodos_mql"
```
kraenhansen, line 26:
```suggestion
export { useTodos } from "./useTodos_mql"
```

web-js/src/hooks/useTodos.js
```diff
@@ -0,0 +1,31 @@
...
+// :state-uncomment-end:
+// :state-uncomment-start: prod-graphql
+import { useTodos } from "./useTodos_graphql"
```
kraenhansen, line 29:
```suggestion
export { useTodos } from "./useTodos_graphql"
```

web-js/src/hooks/useTodos_graphql.js
```diff
@@ -0,0 +1,197 @@
...
+    }
+    return new ApolloClient({
+      link: new HttpLink({
+        uri: graphqlUri,
+        // We define a custom fetch handler for the Apollo client that lets us authenticate GraphQL requests.
+        // The function intercepts every Apollo HTTP request and adds an Authorization header with a valid
+        // access token before sending the request.
+        fetch: async (uri, options) => {
+          const accessToken = await getValidAccessToken();
+          options.headers.Authorization = `Bearer ${accessToken}`;
```
kraenhansen, line 38: This (still) hurts my eyes ... we should really provide a package for this.

web-js/src/components/RealmApp.js
kraenhansen: Honestly, I can see the validity in both arguments.
The root cause of this, is most likely that the Realm Web API is too simple to be really useful without introducing bugs from components not rendering when a user logs in / out.

I see this component as a workaround for an API which is missing important features. This makes it harder for users to consume the app template, but at least we're not teaching users a fast route to trouble.

web-js/public/index.html
```diff
@@ -0,0 +1,43 @@
...
+  <head>
+    <meta charset="utf-8" />
+    <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta name="theme-color" content="#000000" />
+    <meta
+      name="description"
+      content="Web site created using create-react-app"
+    />
+    <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
```
kraenhansen, line 12: Would it be worth it to create a slick logo for this?

kraenhansen (approved):
Nothing major, just a few optional suggestions.

web-js/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+```shell
+REACT_APP_API_TYPE="graphql"
+# REACT_APP_API_TYPE="mql"
+```
+
+> 💡 **Restart to Change Environment Values**
+> The app only sources `.env` values once. If you change a value while the app is running, the app continues to use the value that was defined when it started.
+
+### 3. Specify Your Realm App ID
```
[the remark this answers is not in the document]
nlarew: I originally had them there actually, but the [spec](https://docs.google.com/document/d/18vFW-ATTH7W2S-8BM3vZj5WGqWc66jPV_UoR25bPW6Y/edit) has JS using a `realm.json` file (iOS uses `Realm.plist`) so that the backend server can inject the user's App ID & URL in a platform-agnostic way.

web-js/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "web-vitals": "^1.0.1"
+  },
+  "scripts": {
+    "start": "react-scripts start",
+    "build": "react-scripts build",
+    "test": "react-scripts test",
+    "eject": "react-scripts eject",
+    "generate:prod-mql": "npx bluehawk copy . --destination=../web-mql --state=prod-mql",
+    "generate:prod-graphql": "npx bluehawk copy . --destination=../web-graphql --state=prod-graphql",
+    "generate:prod": "npm run generate:prod-mql && npm run generate:prod-graphql"
```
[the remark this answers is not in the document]
nlarew: Good idea, we can bluehawk these out too :)

web-js/src/hooks/useTodos_graphql.js
nlarew: Agreed! It'd be great to have fetch handlers for common GraphQL clients (e.g. [Apollo](https://www.apollographql.com/docs/react/), [urql](https://formidable.com/open-source/urql/docs/), [graphql-hooks](https://github.com/nearform/graphql-hooks)), that can hook into the SDK's built-in access token refresh handler logic.

web-js/public/index.html
nlarew: Not sure how big of a use case there'll be for saving to an iOS homescreen but we could definitely use a nice favicon and then reuse it as the app icon. I'll ask design to see what's possible.

web-js/src/components/RealmApp.js
nlarew: Not strictly necessary, but we have no reason to hold on to the user's data and it demonstrates another method so sure thing!

web-js/src/components/AppName.js
```diff
@@ -0,0 +1,33 @@
...
+    `Invalid REACT_APP_API_TYPE: "${API_TYPE}". Specifiy "graphql", "mql", or "local" instead.`
+  );
+}
+// :state-end:
+// :state-uncomment-start: prod-mql
+// export const API_TYPE_NAME = "MQL"
+// :state-uncomment-end:
+// :state-uncomment-start: prod-graphql
+// export const API_TYPE_NAME = "GraphQL"
+// :state-uncomment-end:
```
takameyer, line 25: What is this?

takameyer, line 25: What is this?

web-js/src/components/RealmApp.js
takameyer: Just putting my thoughts here.  I agree the provider pattern is the way to go here, but I also think we should be provided from `realm-js` so that the developer doesn't have to write it themselves.

web-js/src/hooks/useCollection.js
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import { useRealmApp } from "../components/RealmApp";
+
+/**
+ * Returns a MongoDB Collection client object
+ * @template DocType extends Realm.Services.MongoDB
+ * @param {Object} config - A description of the collection.
+ * @param {string} [config.service="mongodb-atlas"] - The service name of the collection's linked cluster.
```
takameyer, line 8: What else could this be?  Is it possible to link this to another service?

web-js/src/hooks/useDraftTodos.js
```diff
@@ -0,0 +1,40 @@
...
+    const draftTodo = {
+      _id: new Realm.BSON.ObjectID(),
+      summary: "",
+      isComplete: false,
+    };
+    setDrafts((d) => [...d, draftTodo]);
+  };
+
+  const setDraftTodoSummary = (draft, summary) => {
+    setDrafts((oldDrafts) => {
```
takameyer, line 17: TIL you can give setState a function to derive the new state from the previous state.  Very cool

web-js/src/hooks/useWatch.js
```diff
@@ -0,0 +1,74 @@
...
+ * Opens/manages a change stream and calls provided change handlers for a given MongoDB collection.
+ * @param {Realm.Services.MongoDB.MongoDBCollection<T>} collection - A MongoDB collection client object.
+ * @param {Object} changeHandlers - A set of callback functions to call in response to changes.
+ * @param {(change: Realm.Services.MongoDB.InsertEvent<T>) => void} [changeHandlers.onInsert] - A change handler callback that receives an Insert event.
+ * @param {(change: Realm.Services.MongoDB.UpdateEvent<T>) => void} [changeHandlers.onUpdate] - A change handler callback that receives an Update event.
+ * @param {(change: Realm.Services.MongoDB.ReplaceEvent<T>) => void} [changeHandlers.onReplace] - A change handler callback that receives a Replace event.
+ * @param {(change: Realm.Services.MongoDB.DeleteEvent<T>) => void} [changeHandlers.onDelete] - A change handler callback that receives a Delete event.
+ */
+export function useWatch(collection, changeHandlers) {
+  const filter = React.useMemo(() => ({}), []);
```
takameyer, line 23: Is the filter needed here?

web-js/src/hooks/useTodos_graphql.js
```diff
@@ -0,0 +1,197 @@
...
+  }, [graphql]);
+
+  // Use a MongoDB change stream to reactively update state when operations succeed
+  const taskCollection = useCollection({
+    service: "mongodb-atlas",
+    db: "todo",
+    collection: "Task",
+  });
+  useWatch(taskCollection, {
+    onInsert: (change) => {
```
takameyer, line 81: Do we offer `subscriptions` in our graphql API?  Seems like the GraphQL implementation here is a hybrid of MQL and GraphQL given this collection watcher.

web-js/README.md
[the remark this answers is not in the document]
kraenhansen: Oh okay 👍  makes sense then.

takameyer (approved):
This is very well written.  But it does make me realise how much boilerplate code a Realm developer would have to write to implement a web solution.  I can imagine with a more complex use case many different data models that this would get quite bulky.
We will have to think about ways we can abstract some of this away for our users.  Perhaps by offering our own ApolloProvider/MqlProvider/RealmProviders with generic change handlers.

web-js/src/components/AppName.js
kraenhansen: [bluehawk](https://www.npmjs.com/package/bluehawk) commands 🙂

web-js/src/hooks/useCollection.js
kraenhansen: It's possible to link a MongoDB Realm app to multiple MongoDB Altas clusers as "data sources" and it's therefore possible for a user to define the service name. "mongodb-atlas" is just the default.

cbush (approved):

PR #11, opened 2021-04-05
DOCSP-14518 Android - Implement item deletion

jenniferpeshansky:
WIP, currently the three dots just functions as a delete button rather than generating a popup menu

### DIFF
--- a/android-kotlin/app/src/main/java/com/mongodb/app/ItemAdapter.kt
+++ b/android-kotlin/app/src/main/java/com/mongodb/app/ItemAdapter.kt
@@ -28,19 +28,24 @@ internal class ItemAdapter(data: OrderedRealmCollection<Item>, val user: io.real
         holder.name.text = obj?.name
 
         holder.itemView.setOnClickListener {
+            deleteItem(holder.id!!)
+            //TODO: implement popup menu with delete option rather than just deleting
+            /*
             run {
                 val popup = PopupMenu(holder.itemView.context, holder.menu)
                 val menu = popup.menu
-                menu.add(0, -1  , Menu.NONE, "Delete Item")
+                val deleteCode = -1
+                menu.add(0, deleteCode, Menu.NONE, "Delete Item")
                 popup.setOnMenuItemClickListener { menuItem: MenuItem? ->
                     when (menuItem!!.itemId) {
-                        -1 -> {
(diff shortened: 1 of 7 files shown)


android-kotlin/app/src/main/res/values/strings.xml
```diff
@@ -8,4 +8,5 @@
     <string name="already_have_account">Already have an account? Log in</string>
     <string name="does_not_have_account">Don\'t have an account? Create account</string>
     <string name="log_out">Log Out</string>
+    <string name="more">\u22EE</string>
```
nathan-contino-mongo, line 11: If we plan on popping up a menu here, this is fine, but if we don't have any other interaction plans for this button we should just change the icon to a trash can or an X, something that makes it obviously a delete interaction. particularly because deleting the item is NOT what most users would expect from a triple-dot menu.

jenniferpeshansky: true, I was planning to change it to an X if I couldn't get the pop up menu to work

nathan-contino-mongo (approved):
LGTM

android-kotlin/app/src/main/java/com/mongodb/app/ItemAdapter.kt
```diff
@@ -17,17 +16,52 @@ import org.bson.types.ObjectId
...
+            }
+            popup.show()
+        }
+    }
+
+    /**
+     *  Allows the user to delete an item from the realm.
+     */
+    private fun deleteItem(id: ObjectId) {
+        val config = SyncConfiguration.Builder(user, partition).build()
```
cbush, line 51: I wonder if the correct pattern would actually be to set the default realm configuration when you log in. Then you can get the default realm from here without needing to know the user. Another option would be to pass in the configuration rather than the user and partition. That would make this adapter more flexible.

cbush (approved):

android-kotlin/app/src/main/java/com/mongodb/app/ItemAdapter.kt
nathan-contino-mongo: Yeah, setting the default realm in somewhere in this project makes a lot of sense, since generally you're only using one realm in this app. +1.

PR #12, opened 2021-04-06
DOCSP-14519 Android - Implement check/uncheck item functionality

jenniferpeshansky:


android-kotlin/app/src/main/java/com/mongodb/app/ItemAdapter.kt
```diff
@@ -32,11 +35,24 @@ internal class ItemAdapter(data: OrderedRealmCollection<Item>, private val confi
         val obj: Item? = getItem(position)
         holder.id = obj?._id
         holder.name.text = obj?.name
+        holder.checkbox.isChecked = obj?.checked?:false
+
+        /**
+         *  Allows a user to check and uncheck an item and updates its status in the realm.
```
cbush, line 41: I would prefer line comment syntax (not javadoc) within methods.

android-kotlin/app/src/main/java/com/mongodb/app/ItemsActivity.kt
```diff
@@ -123,6 +124,20 @@ class ItemsActivity : AppCompatActivity() {
         input.requestFocus()
     }
 
+/*    fun onCheckboxClicked(view: View) {
```
cbush, line 127: Is this needed?

cbush (approved):
A few comments.

android-kotlin/app/src/main/java/com/mongodb/app/ItemsActivity.kt
jenniferpeshansky: nope

android-kotlin/app/src/main/java/com/mongodb/app/ItemAdapter.kt
jenniferpeshansky: extracted the methods and moved the comments

android-kotlin/app/src/main/java/com/mongodb/app/ItemAdapter.kt
```diff
@@ -32,26 +35,44 @@ internal class ItemAdapter(data: OrderedRealmCollection<Item>, private val confi
         val obj: Item? = getItem(position)
         holder.id = obj?._id
         holder.name.text = obj?.name
+        holder.checkbox.isChecked = obj?.checked?:false
+        holder.checkbox.setOnClickListener { onCheckboxClicked(holder) }
+        holder.menu.setOnClickListener { onMenuClicked(holder) }
```
jenniferpeshansky, line 40: I'm not sure if one-lining these like this is bad practice or not, it looks nicer to me but I'm open to changing it

cbush: That's fine.

android-kotlin/app/src/main/java/com/mongodb/app/TaskActivity.kt
```diff
@@ -19,28 +20,43 @@ import io.realm.RealmConfiguration
...
 
-        recyclerView = findViewById(R.id.items_list)
+        recyclerView = findViewById(R.id.task_list)
+        recyclerView.layoutManager = LinearLayoutManager(this@TaskActivity)
+        recyclerView.adapter = object : RecyclerView.Adapter<RecyclerView.ViewHolder>() {
+            override fun onCreateViewHolder(
+                parent: ViewGroup,
+                viewType: Int
+            ): RecyclerView.ViewHolder {
+                TODO("Not yet implemented")
```
cbush, line 45: This might be a bit cleaner. I don't think this method can ever be called, but then we're leaving TODOs in people's code. Another option would be to just `throw` in this method.
```            override fun onCreateViewHolder(
                parent: ViewGroup,
                viewType: Int
            ): RecyclerView.ViewHolder {
                // This can never be called
                return object: RecyclerView.ViewHolder(parent.rootView) {}
            }
```

cbush (approved):

PR #13, opened 2021-04-09
Android Template App

jenniferpeshansky:


android-kotlin/.gitignore
```diff
@@ -0,0 +1,16 @@
+*.iml
+.gradle
+/local.properties
+/.idea/caches
```
cmelchior, line 4: All of these shouldn't be needed if you have `.idea`

android-kotlin/app/.gitignore
```diff
@@ -0,0 +1,2 @@
+/build
```
cmelchior, line 1: Normally we try to just have the top-level .gitignore file, since that makes it easier to update.

android-kotlin/app/build.gradle
```diff
@@ -0,0 +1,54 @@
...
+        jvmTarget = '1.8'
+    }
+}
+
+realm {
+    syncEnabled = true
+}
+
+dependencies {
+
```
cmelchior, line 41:
```suggestion
```

android-kotlin/app/src/main/java/com/mongodb/app/RealmApp.kt
```diff
@@ -0,0 +1,37 @@
...
+    override fun onCreate() {
+        super.onCreate()
+        Realm.init(this)
+         realmApp = App(
+            AppConfiguration.Builder(getString(R.string.mongodb_realm_app_id))
+                .build())
+
+        // Enable more logging in debug mode
+        if (BuildConfig.DEBUG) {
+            RealmLog.setLevel(LogLevel.ALL)
```
cmelchior, line 32: This is very verbose and result in a A LOT of log output which might be confusing...for a starter experience I would suggest using `LogLevel.DEBUG`

android-kotlin/app/src/main/java/com/mongodb/app/RealmApp.kt
```diff
@@ -0,0 +1,37 @@
...
+
+/*
+*  Sets up the Realm App and enables Realm-specific logging in debug mode.
+*/
+class App: Application() {
+
+    override fun onCreate() {
+        super.onCreate()
+        Realm.init(this)
+         realmApp = App(
```
cmelchior, line 26: This can be replaced with just `App(getString(R.string.mongodb_realm_app_id))`

android-kotlin/app/src/main/java/com/mongodb/app/Task.kt
```diff
@@ -0,0 +1,12 @@
+package com.mongodb.app
+
+import io.realm.RealmObject
+import io.realm.annotations.PrimaryKey
+import org.bson.types.ObjectId
+
+open class Task(
+        @PrimaryKey var _id: ObjectId = ObjectId(),
```
cmelchior, line 8: I would rename this to 

```
@RealmField("_id")
@PrimaryKey
var id: ObjectId = ObjectId()
```

For a more idiomatic name in Kotlin

android-kotlin/app/src/main/java/com/mongodb/app/TaskActivity.kt
```diff
@@ -0,0 +1,136 @@
...
+        if (user == null) {
+            startActivity(Intent(this, LoginActivity::class.java))
+        }
+        else {
+            val partition = user.id
+            config = SyncConfiguration.Builder(user, partition).build()
+            Realm.getInstanceAsync(config, object : Realm.Callback() {
+                override fun onSuccess(realm: Realm) {
+                    this@TaskActivity.userRealm = realm
+                    recyclerView.adapter = TaskAdapter(realm.where<Task>().sort("_id").findAll(), config)
```
cmelchior, line 61: I would use findAllAsync() here, otherwise the query runs on the UI thread which can slow it down.

android-kotlin/app/src/main/java/com/mongodb/app/TaskActivity.kt
```diff
@@ -0,0 +1,136 @@
...
+    private fun onFabClicked() {
+        val input = EditText(this)
+        val dialogBuilder = AlertDialog.Builder(this)
+        dialogBuilder.setMessage("Enter task name:")
+                .setCancelable(true)
+                .setPositiveButton("Create") { dialog, _ ->
+                    run {
+                        dialog.dismiss()
+                        val task = Task()
+                        task._partition = realmApp.currentUser()!!.id
```
cmelchior, line 110: This isn't needed... It is inserted automatically by the protocol

android-kotlin/app/src/main/java/com/mongodb/app/Task.kt
```diff
@@ -0,0 +1,12 @@
...
+        var _partition: String = "",
```
cmelchior, line 9: I assume this is the partition key. If so, this isn't needed... It is inserted automatically by the protocol

android-kotlin/app/src/main/java/com/mongodb/app/TaskAdapter.kt
```diff
@@ -0,0 +1,75 @@
...
+        holder.checkbox.isChecked = task.isComplete
+        holder.checkbox.setOnClickListener { onCheckboxClicked(holder) }
+        holder.menu.setOnClickListener { onMenuClicked(holder) }
+    }
+
+    /**
+     *  Allows a user to check and uncheck a task and updates its status in the realm.
+     */
+    private fun onCheckboxClicked(holder: TaskViewHolder) {
+        val realm: Realm = Realm.getInstance(config)
```
cmelchior, line 46: There shouldn't be a need to open the Realm here... you can get it from the RealmResults using `results.getRealm()`

android-kotlin/app/src/main/java/com/mongodb/app/TaskAdapter.kt
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     *  Creates a popup menu that allows the user to delete a task from the realm.
+     */
+    private fun onMenuClicked(holder: TaskViewHolder) {
+        val popup = PopupMenu(holder.itemView.context, holder.menu)
+        popup.menu.add(0, R.id.action_delete, Menu.NONE, "Delete")
+        popup.setOnMenuItemClickListener { menuItem: MenuItem ->
+            when (menuItem.itemId) {
+                R.id.action_delete -> {
+                    val realm: Realm = Realm.getInstance(config)
```
cmelchior, line 63: Same

android-kotlin/app/src/main/res/layout/activity_login.xml
```diff
@@ -0,0 +1,124 @@
...
+        android:textColor="@color/blue"
+        app:backgroundTint="@color/white"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@+id/button_create_login"
+        app:layout_constraintVertical_bias="0.091"
+        style="?android:attr/borderlessButtonStyle" />
+
+    <ProgressBar
```
cmelchior, line 97: This doesn't appear to be used anywhere?

android-kotlin/app/src/test/java/com/mongodb/app/ExampleUnitTest.kt
```diff
@@ -0,0 +1,17 @@
...
+import org.junit.Test
+
+import org.junit.Assert.*
+
+/**
+ * Example local unit test, which will execute on the development machine (host).
+ *
+ * See [testing documentation](http://d.android.com/tools/testing).
+ */
+class ExampleUnitTest {
```
cmelchior, line 12: Any reason for this class? Maybe just remove it?

android-kotlin/app/src/androidTest/java/com/mongodb/app/ExampleInstrumentedTest.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import org.junit.Assert.*
+
+/**
+ * Instrumented test, which will execute on an Android device.
+ *
+ * See [testing documentation](http://d.android.com/tools/testing).
+ */
+@RunWith(AndroidJUnit4::class)
+class ExampleInstrumentedTest {
```
cmelchior, line 17: Any reason for this class? Maybe just remove it?

cmelchior (approved):
Mostly just minor comments. Otherwise looking good 👍

android-kotlin/app/src/main/java/com/mongodb/app/TaskAdapter.kt
nathan-contino-mongo: I don't think this is true since the adapter actually accepts an OrderedRealmCollection as an argument, and that class doesn't have a `getRealm()` method that I can see. We could always accept a RealmResults instead, but then folks can't use other implementations of OrderedRealmCollection... considering the fact that initializing a realm that's already open is quite cheap, I think we might just have to accept this one.

android-kotlin/app/src/main/java/com/mongodb/app/TaskAdapter.kt
nathan-contino-mongo: same as above

nathan-contino-mongo:
@cmelchior One question for you, otherwise took your suggestions exactly.

cbush (approved):

PR #15, opened 2021-05-20
Add baseUrl to Realm.plist and use it in Swift app

cbush:


ios-swift/App/Realm.plist
```diff
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>baseUrl</key>
```
nlarew, line 5: nit: should we make this consistent across the SDKs? The [spec](https://docs.google.com/document/d/1gmbjUjeSULW4aI3OV9ytBqU0sypQ78H2cpIKIHr1xWM/edit#) has some using `base-url` and others using just `url` (none currently say `baseUrl` though it's fine if we choose that)

ios-swift/App/Realm.plist
```diff
@@ -2,6 +2,8 @@
...
+	<string>https://realm.mongodb.com</string>
```
nlarew, line 6: I don't think we use this outside of the GraphQL app, but it should also include the project/app id in the path since you can't access the Realm UI with only the App ID™️.

e.g. `https://realm.mongodb.com/groups/57879f6ce4b04cba440ba8c5/apps/5b366d043cd49d6e3add621e/`

nlarew (approved):

ios-swift/App/Realm.plist
nlarew: I'm actually a bit confused here - the spec lists both a `base-url` and `app-url` for iOS and Android but only a `url` for web 🤔

ios-swift/App/realmApp.swift
```diff
@@ -12,5 +12,7 @@ var realmApp: App = {
     let data = NSData(contentsOfFile: path)! as Data
     let realmPropertyList = try! PropertyListSerialization.propertyList(from: data, format: nil) as! [String: Any]
     let appId = realmPropertyList["appId"]! as! String
-    return App(id: appId)
+    let baseUrl = realmPropertyList["baseUrl"]! as! String
+    return App(id: appId,
+               configuration: AppConfiguration(baseURL: baseUrl, transport: nil, localAppName: nil, localAppVersion: nil))
```
nlarew, line 17: Do we need to specify this baseUrl? Doesn't it already default to `realm.mongodb.com`?

cbush: Yes but this just allows the devs to use realm-dev or whatever.

ios-swift/App/Realm.plist
cbush: For baseUrl on iOS it's just the domain I believe.

ios-swift/App/Realm.plist
cbush: Discussed in slack - will follow up with Andrew and make consistent across the board.

andrewzlchen: yeah sorry that was my bad on the spec. 

I meant to make all of the clients export the same metadata, but I must've been distracted halfway

PR #18, opened 2021-06-15
Update .NET template to use JSON config format

cbush:
- This format is specified by the backend
- Allow backend URL config


xamarin/_Shared/realm.json
```diff
@@ -0,0 +1,4 @@
+﻿{
+    "app_id": "todo-sync-tmikv",
```
MongoCaleb, line 2: Did you intentionally include an appID here?

MongoCaleb (approved):
LGTM

xamarin/_Shared/realm.json
cbush: Yes. Will be overridden by backend generation.

PR #19, opened 2021-06-15
Update RN config to use format

cbush:
- Format specified by backend uses snake_case
- Fix bug where baseUrl was only loaded on one component


react-native/realm.json
```diff
@@ -1,3 +1,4 @@
 {
-  "appId": "<Your App ID>"
+  "app_id": "todo-sync-tmikv",
```
mongomoe, line 2: Did you mean to include your app id?

react-native/src/RealmApp.js
```diff
@@ -0,0 +1,15 @@
...
+if (!base_url) {
+  throw 'Missing Realm base URL. Set base_url in realm.json';
+}
+
+const appConfiguration = {
+  id: app_id,
+  baseUrl: base_url
+};
+
+export const realmApp = new Realm.App(appConfiguration);
```
mongomoe, line 15: why create a realmApp export when we have the in built SDK method - https://docs.mongodb.com/realm-sdks/js/latest/Realm.App.html#getApp?

cbush: That would require loading the configuration in various places as well. Note we're not just using the app ID.

react-native/realm.json
cbush: Yes, it makes it easier for us to test and will be overridden by the generator.

react-native/src/RealmApp.js
mongomoe: getApp doesn't require the configuration, just the ID. So initially you can specify a configuration that includes an app id and a base URL and then wherever you want to use the realm app, just pass the exported app id to the getApp method

mongomoe: Not saying this approach doesn't work, it definitely does. I think it's a bit redundant since it accomplishes the same thing `getApp()` does, except without passing an ID. 

Especially if our goal is to make it as easy as possible to turn a scaffolded template app into their app, perhaps we don't want to include a function which they might delete later on?

cbush: Ah that's cool. One question: how do we guarantee that any Realm.App.getApp() calls happen after the initialization with the base url?

mongomoe: I see your point, getApp would create a new Realm App with just the app id, rather than throw an error and say that 'the app has not been initialized yet'

mongomoe (approved):
LGTM 👍

PR #21, opened 2021-06-16
Update android to use base_url

cbush:


nathan-contino-mongo (approved):
LGTM -- just tested, and this works for me!

PR #25, opened 2021-06-24
Update todo-sync backend export to use CLI v2

cbush:


todo-sync/data_sources/mongodb-atlas/config.json
```diff
@@ -0,0 +1,10 @@
+{
+    "name": "mongodb-atlas",
+    "type": "mongodb-atlas",
+    "config": {
+        "clusterName": "Cluster0",
```
nlarew, line 5: Just thought about this - will Realm handle cluster names on the server side when it creates the template app instance? What if they use a name other than Cluster0?

nlarew (approved):

todo-sync/data_sources/mongodb-atlas/config.json
cbush: @andrewzlchen any thoughts?

PR #27, opened 2021-06-25
Xamarin quick fixes

cbush:
- Fixed: When starting the app while already logged in, and then logging out, the app would not close the existing task view. Upon re-log in, crash due to realm already being opened. Also, pushing the login view is incorrect here as a second logout would then pop to root, which would have been the first task view.
- Unified font size between sign up/sign in views for less distraction.
- Fixed: When logging out and returning to the login view, the isLoggingIn could be set to the incorrect state.


MongoCaleb (approved):
LGTM

PR #29, opened 2021-07-02
[Web - GraphQL/MQL] QA

nlarew:
Fixes small integration issues we uncovered in the QA session

- Lowercase collection name
- Remove extra console.log
- Use dynamic baseUrl for GraphQL

### DIFF
--- a/generated/web-graphql/src/hooks/useTodos_mql.js
+++ b/generated/web-graphql/src/hooks/useTodos_mql.js
@@ -21,7 +21,7 @@ export function useTodos() {
   const taskCollection = useCollection({
     clusterName,
     db: "todo",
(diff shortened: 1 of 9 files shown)


cbush (approved):

generated/web-graphql/src/hooks/useTodos_mql.js
```diff
@@ -21,7 +21,7 @@ export function useTodos() {
   const taskCollection = useCollection({
     clusterName,
     db: "todo",
-    collection: "Task",
+    collection: "task",
```
andrewzlchen, line 24: hmm this broke the client as well. can we make the same change for the backend?

PR #38, opened 2021-08-09
Fix null reference exception on any login alert

cbush:
- Removed current user check since this method is only ever called in the context where
current user is definitely null
- Move unsafe code into try block


dacharyc (approved):
This looks reasonable 👍

nathan-contino-mongo (approved):
Combining my xamarin knowledge with Dachary's to add up to (hopefully) one legitimately semi-informed LGTM

PR #39, opened 2021-08-10
Use LOCAL deployment instead of GLOBAL

nlarew:


cbush:
Discussed in slack - closing for now

PR #44, opened 2021-12-07
Electron auth and routing

mongodben:


electron/src/renderer/components/Context.jsx
```diff
@@ -0,0 +1,5 @@
+import { createContext } from 'react';
+
+const Context = createContext({});
```
mongomoe, line 3: does this really need to be its own file? Especially since this is just a generic createContext and only used in 1 place (App.jsx)

electron/src/renderer/App.jsx
```diff
@@ -0,0 +1,104 @@
...
+      <Context.Provider value={{ onLogIn, onSignUp, isLoggedIn }}>
+        <div>
+          <Link to="/">
+            <h1>Realm Todo</h1>
+          </Link>
+          {isLoggedIn && <button onClick={onLogOut}>Log Out</button>}
+        </div>
+        <Switch>
+          <Route path="/" component={Welcome} exact />
+          <Route path="/log-in" component={LogIn} />
```
mongomoe, line 87: this is fine for now, but we really should consolidate /log-in and /sign-up to be one single route

electron/src/renderer/App.jsx
```diff
@@ -0,0 +1,104 @@
...
+  MemoryRouter as Router,
+  Switch,
+  Route,
+  Link,
+  Redirect,
+  useHistory,
+  useLocation,
+} from 'react-router-dom';
+import './App.css';
+import { realmApp, Authentication } from './realm';
```
mongomoe, line 12: remember to replace realmApp with the inbuilt getApp

electron/src/renderer/components/Context.jsx
mongodben: used in a bunch of files, anywhere that the `useContext` hook is imported. 4 files as of now.

electron/src/renderer/App.jsx
mongodben: after reviewing the [method docs for `getApp()`](https://docs.mongodb.com/realm-sdks/js/latest/Realm.App.html#getApp) i'm not convinced that it would be better to use `getApp` instead of the current implementation. this is b.c the `getApp()` documentation requires passing in the app id as an argument every time it's called. 

so to use the `getApp()` approach, we'd need to either: 
1. define APP_ID as an environmental variable or variable to be imported, and pass it to every invocation of `getApp`
2. use the string of the App ID in every invocation, which would not be a very DRY solution

lmk what you think here

electron/src/renderer/App.jsx
```diff
@@ -0,0 +1,104 @@
...
+  );
+};
+
+const App = () => {
+  const history = useHistory();
+  const [isLoggedIn, setIsLoggedIn] = useState(false);
+
+  useEffect(() => {
+    console.log(realmApp.currentUser?.isLoggedIn);
+    setIsLoggedIn(!!realmApp.currentUser?.isLoggedIn);
```
mongomoe, line 41: personally not a big fan of using the double exclamation mark (!! or not not), especially here since the templates are intended to appeal to the broadest audience. A simpler alternative is something like:

`` app.currentUser?.isLoggedIn == true? setIsLoggedIn(true) : setIsLoggedIn(false)``

in the above example, if app.currentUser?.isLoggedIn == undefined, then the program will call setIsLoggedIn(false)

mongomoe (approved):
lgtm w/ a couple thoughts

electron/src/renderer/components/Context.jsx
mongomoe: oh i see makes sense, thanks for clarifying

electron/src/renderer/App.jsx
mongomoe: Fair point on it not being as DRY, but we should define the App ID in a config file (not env). We can take the standard set out in the web & rn templates, of using realm.json: https://github.com/mongodb-university/realm-template-apps/blob/main/react-native/realm.json

electron/src/renderer/App.jsx
mongodben: > in the above example, if app.currentUser?.isLoggedIn == undefined, then the program will call setIsLoggedIn(false)
yes, and this is correct behavior 

anyways, can refactor to use ternary for readability

mongodben:
note a bit of git funkiness w deleted files reflected in last 3 commits in history

PR #45, opened 2021-12-08
Add todos functionality and UI

mongodben:
add CRUD functionality for Todos w sync to MongoDB Realm. 

@mongomoe pls try to be pretty thorough in this PR since i have absolutely no confidence this is the correct approach, though it seems to work well, as far as i can tell. 

we might wanna get nick/JS team to review as well.


mongodben:
closing to fix some git funkiness, where some rebased commits are here when they shouldn't be

PR #52, opened 2022-04-08
migrate to flexible sync (DOCSP-21836)

MongoCaleb:
Updates the existing subscription-based template app to use Flexible Sync. 

[Jira](https://jira.mongodb.org/browse/DOCSP-21836)


cbush:
Sorry this wasn't clear in the ticket: this should be a net new app - so I would copy the directory to `xamarin-flex-sync` and go from there.
 
We also need the backend exported. That also should be a sibling to the PBS backend.

cbush:
If you want to be fancy, you can use bluehawk states for this and have the output in the repo's `generated/` folder.

generated/README.md
```diff
@@ -0,0 +1,46 @@
+﻿# Xamarin C# Template App
```
cbush, line 1: This file does not belong in `generated/`

xamarin/realm-todo-dotnet/App.xaml.cs
```diff
@@ -44,24 +54,28 @@ protected override void OnStart()
             }
         }
 
-        private Realms.Sync.AppConfiguration LoadAppConfiguration()
+        private bool GetAppConfiguration()
```
cbush, line 57: The naming here, at least, is questionable. "Get" implies it would return the app configuration. I think "HasAppConfiguration()" would be a better name for something returning a bool. However, why do we need this check? Since it's loading into member variables I think the best approach here would be leaving it as "LoadAppConfiguration", making it `void` and removing the try/catch block that just rethrows anyway.

cbush (commented):
One thing to update is the manifest file in the root. The "old" app should point to the new location.

xamarin/realm-todo-dotnet/App.xaml.cs
```diff
@@ -44,21 +44,15 @@ protected override void OnStart()
             }
         }
 
-        private Realms.Sync.AppConfiguration LoadAppConfiguration()
+        private string LoadAppConfiguration()
```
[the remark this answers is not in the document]
MongoCaleb: (as we do in the catch statement immediately above)

README.md
```diff
@@ -2,3 +2,64 @@
...
+   partition-based app), consider using the Bluehawk `state` tags and using
+   Bluehawk to copy each app to its own subdirectory within the `/generated`
+   directory.
+
+2. Be sure your project uses a `realm.json` file to get the app id and
+   base url info:
+
+   ```
+   {
+      "appId": "todo-sync-jxgjv",
```
mongomoe, line 20: should this app Id be in the Readme? Or maybe just appId: "your-app-id"

generated/xamarin-flexible-sync/realm-todo-dotnet/App.xaml.cs
```diff
@@ -0,0 +1,75 @@
...
+            catch (Exception e)
+            {
+                // A NullReferenceException occurs if:
+                // 1. the config file does not exist, or
+                // 2. the config does not contain an "appId" or "baseUrl" element.
+
+                // If the appId value is incorrect, we handle that
+                // exception in the Login page.
+
+                throw e;
```
mongomoe, line 42: Let's add some kind of error message. Maybe just print a simple line explaining the possible errors here

generated/xamarin-flexible-sync/realm-todo-dotnet/App.xaml.cs
```diff
@@ -0,0 +1,75 @@
...
+                   GetManifestResourceStream("RealmTemplateApp." + "realm.json"))
+                using (StreamReader reader = new StreamReader(stream))
+                {
+                    var json = reader.ReadToEnd();
+                    var parsedJson = JsonConvert.DeserializeObject<Dictionary<string, string>>(json);
+                    appId = parsedJson["appId"];
+                    baseUrl = parsedJson["baseUrl"];
+                }
+            }
+            catch (Exception ex)
```
mongomoe, line 60: Nitpick: In some places you have `Exception ex` and others `Exception e`, can you make this consistent?

generated/xamarin-partition-based-sync/realm-todo-dotnet/TaskPage.xaml.cs
```diff
@@ -0,0 +1,153 @@
...
+    {
+        private Realm taskRealm;
+        private User user;
+        private IEnumerable<Task> _tasks;
+
+        public TaskPage()
+        {
+            InitializeComponent();
+            user = App.RealmApp.CurrentUser;
+            var config = new SyncConfiguration(user.Id.ToString(), user);
```
mongomoe, line 23: SyncConfiguration should be PartitionSyncConfiguration

xamarin/realm-todo-dotnet/TaskPage.xaml.cs
```diff
@@ -20,10 +20,30 @@ public TaskPage()
         {
             InitializeComponent();
             user = App.RealmApp.CurrentUser;
+            // :state-start: partition-based-sync
             var config = new SyncConfiguration(user.Id.ToString(), user);
```
mongomoe, line 24: SyncConfiguration should be PartitionSyncConfiguration

papafe:
I'm a little bit confused by the structure of the PR. `realm-todo` seems to be in several folders?

generated/xamarin-flexible-sync/realm-todo-dotnet/TaskPage.xaml.cs
```diff
@@ -0,0 +1,164 @@
...
+            }
+            catch (Exception ex)
+            {
+                await DisplayAlert("Error Fetching Tasks", ex.Message, "OK");
+            }
+        }
+
+        private void SetUpTaskList()
+        {
+            if (_tasks == null || _tasks.Count() == 0)
```
papafe, line 54: I'm not sure how useful is this part. The loading does not happen when we call `taskRealm.All`.

If you want to do it you should do it when calling `WaitForSynchronizationAsync`

generated/xamarin-flexible-sync/realm-todo-dotnet/TaskPage.xaml.cs
```diff
@@ -0,0 +1,164 @@
...
+                IsComplete = false
+            };
+
+            taskRealm.Write(() =>
+            {
+                taskRealm.Add(newTask);
+            });
+        }
+
+        private void chkCompleted_PropertyChanged(object sender, PropertyChangedEventArgs e)
```
papafe, line 86: Methods should be PascalCase

generated/xamarin-flexible-sync/realm-todo-dotnet/realm.json
```diff
@@ -0,0 +1,4 @@
+﻿{
+  "appId": "todo-sync-jxgjv",
```
papafe, line 2: Do we want to let the users use our server?

papafe (changes requested):

README.md
MongoCaleb: Supposed to be here for template apps. There's a script that parses things and updates the app id, too.

generated/xamarin-flexible-sync/realm-todo-dotnet/realm.json
MongoCaleb: Yes; it gets updated by a script before deploying to docs.

MongoCaleb:
@papafe - it *is* weird. That's because we use a single solution to generate two different solutions -- one for partition-based sync and one for flexible sync. So what you're seeing in this PR is the "main" solution and each of the generated solutions.

generated/xamarin-flexible-sync/realm-todo-dotnet/TaskPage.xaml.cs
```diff
@@ -53,9 +55,7 @@ private void SetUpTaskList()
         {
             if (_tasks == null || _tasks.Count() == 0)
```
papafe, line 56: If `_tasks==null` it makes sense to initialize it, but if `Count() == 0`, then it doesn't. If there are no tasks, then there are not tasks, assigning it again will not change it

generated/xamarin-partition-based-sync/realm-todo-dotnet/App.xaml.cs
```diff
@@ -34,7 +35,7 @@ protected override void OnStart()
                 NavigationPage.SetHasBackButton(navPage, false);
                 MainPage = navPage;
             }
-            catch (Exception e)
+            catch (Exception ex)
             {
                 // A NullReferenceException occurs if:
```
papafe, line 40: Given that here you are capturing all exceptions, and not only the `NullReference` ones, I would change the text to say `An exception can occur if`. 
It could be that different exceptions are coming from that piece of code anyways, so we stay more on the general side

generated/xamarin-flexible-sync/realm-todo-dotnet/TaskPage.xaml.cs
```diff
@@ -83,19 +83,6 @@ private async void New_Button_Clicked(object sender, EventArgs e)
             });
         }
 
-        private void chkCompleted_PropertyChanged(object sender, PropertyChangedEventArgs e)
```
papafe: I suppose this has been removed because it's done automatically through the bindings, am I correct?

generated/xamarin-flexible-sync/realm-todo-dotnet/TaskPage.xaml.cs
```diff
@@ -0,0 +1,151 @@
...
+               var defaultSubscription = taskRealm.All<Task>()
+                   .Where(t => t.OwnerId == user.Id);
+               subscriptions.Add(defaultSubscription);
+           });
+        }
+        
+        protected override async void OnAppearing()
+        {
+            base.OnAppearing();
+            WaitingLayout.IsVisible = true;
```
papafe, line 43: Is this set to false somewhere?

papafe (commented):
@MongoCaleb Thanks for the clarification! So I took a look mostly at the code in  `generated/xamarin-flexible-sync`.

mongomoe (approved):
LGTM 👍

PR #54, opened 2022-05-03
REALMC-13096: add generic sync template to manifest

jandersongo:


cbush:
Hi, what's this?

manifest.json
```diff
@@ -1,4 +1,12 @@
 {
+  "sync.todo": {
+    "name": "Sync Todo App",
+    "repo_owner": "mongodb-university",
+    "repo_name": "realm-template-apps",
+    "backend_path": "todo-sync",
+    "metadata_filename": "Realm",
```
cbush, line 7: If there's no client I don't think we need "metadata_filename" or "file_format". See "Triggers Template" below for an example!

jandersongo:
> Hi, what's this?

Hello! 

I'm on the Realm Cloud team and we're planning to create a 'sync template' category as an alernative to listing each frontend language as options. 
See 'Real-time Sync' in designs here:

image

cbush:
Your change on this repo will allow people to create the todo-specific backend without getting a client. Is that the intention here? Or is this more of a question of grouping the template selections in the UI? cc @andrewzlchen

andrewzlchen:
Yup, its for grouping all of the sync templates together in the UI

cbush:
I don't really see how this change is relevant to UI-side grouping. All this does is allow one to clone the backend without a client. I'm vaguely ok with that, but I'm less ok with approving a PR that doesn't explain how it solves the problem it's targeting.

Why not do some logic on that side for this purpose? For example - if they all share the same backend path they can be grouped.

We could also expand the manifest definition to have groups baked in here, but I think the grouping/layout would be better decided on the UI side.

andrewzlchen:
Yeah, that's definitely valid. That's our bad for not looping you guys in earlier

and yeah, we already do things in the backend that tell us what templates are "compatible" with each other. "Compatibility" amongst templates means that they share the same backend.

in its current state, a template in the manifest is pairing between a frontend and a backend which was a requirement in the TSAs project, but unfortunately, requirements changed and it seems like we have a new definition of what a "template" is.

The changes we're making in this PR is a stopgap-ish change that doesn't make the assumption of what client users are planning to use. 

And to be fair, we don't necessarily need to merge this PR though it'd be nice. We could just silently make all "sync templates" select the swift template by default in the dropdown for frontend language when they try to export their app via the UI modal.


As an aside, with the shift in what the definition of a "template" is, I think the ideal long-term format of the manifest file would be to call the backends "templates" and have a new field for `"clients"` in the manifest.json so that it'd look like this:

```json
{
  "todo-sync": {
    "path": "todo_sync",
    "clients": {
      "react-native": {
        "path": "react-native",
        "metadata_path": "react-native",
        "metadata_filename": "realm",
        "file_format": "json"
      },
      "xamarin": {
         "path": "generated/xamarin-partition-based-sync",
         "metadata_path": "generated/xamarin-partition-based-sync/realm-todo-dotnet",
         "metadata_filename": "realm",
         "file_format": "json" 
      }
    }
  }
}
```

but that's something that we can follow up on as 2-3 improvement tickets

cbush:
Yep, that all sounds good! Like I said - definitely not against this PR but I needed some context before giving my +1.

cbush (approved):

PR #63, opened 2022-05-24
Refactor flex sync

cbush:
- Move logIn and getRealm to reusable
- Remove CLI options, instead read from realm.json


MongoCaleb (approved):
lgtm

PR #64, opened 2022-05-25
Update manifest and zip file name

cbush:
- Use a new zip file name so that we can merge things and update the zip file. When backend team is ready to consume it, they can change the name of the zip file they are looking for.
- Add flex sync versions to manifest file


mongomoe (approved):
LGTM 👍

PR #67, opened 2022-05-31
Update end user facing strings for App Services rename

cbush:
See https://jira.mongodb.org/browse/DOCSP-22864

- In cases where we said "MongoDB Realm Sync" -> "Atlas Device Sync" (per product)
- Otherwise when Sync is not used/mentioned -> "Atlas App Services" (note: Sync is not used at all in GraphQL, etc)


nathan-contino-mongo (approved):
LGTM

PR #68, opened 2022-06-01
Flutter sync

mongodben:
add flexible sync to flutter template app


flutter/lib/components/todo_list.dart
```diff
@@ -29,7 +30,10 @@ class _TodoListState extends State<TodoList> {
 
   @override
   Widget build(BuildContext context) {
-    final realm = Provider.of<Realm>(context);
+    final realm = Provider.of<Realm?>(context);
+    if (realm == null) {
+      return Container();
+    }
```
mongodben, line 36: is there better way to deal w this before realm has been initialized?

flutter/lib/main.dart
```diff
@@ -0,0 +1,58 @@
...
+        update: (context, app, previousRealm) {
+          if (app.currentUser != null) {
+            previousRealm?.close();
+            return initRealm(app.currentUser!);
+          }
+          return null;
+        },
+        dispose: (_, realm) => realm?.close(),
+      )
+    ], child: const App()));
```
mongodben, line 24: did this MultiProvider/ProxyProvider implementation to provide top-level access to both realm and auth, while having the behavior of realm conditional on auth.

needed to put realm top level so that `showModalBottomSheet()` has access to it ([see this issue](https://stackoverflow.com/questions/63818402/flutter-cant-access-provider-from-showmodalbottomsheet))

flutter/lib/components/todo_list.dart
```diff
@@ -4,16 +4,15 @@ import 'package:realm/realm.dart';
 import 'todo_item.dart';
 import 'package:flutter_todo/realm/schemas.dart';
 
-const _animationDuration = Duration(milliseconds: 300);
-
 class TodoViewModel {
   final String id;
   final bool isComplete;
```
desistefanova, line 9: The logic written behind onChanged could be a method or setter SetState in TodoViewModel with a nullable argument. On this way the View will be kept simple as possible.
``` dart
  return Checkbox(
      checkColor: Colors.white,
      fillColor: MaterialStateProperty.resolveWith(getColor),
      value: viewModel.isComplete,
      onChanged: (bool? value) { //This could be moved to TodoViewModel.
        realm.write(() {
          viewModel.todo.isComplete = value ?? false;
        });
      },
    );
```

desistefanova (changes requested):
I have put some comments in relation to MVVM pattern.

flutter/ios/Podfile.lock
```diff
@@ -1,21 +1,27 @@
 PODS:
   - Flutter (1.0.0)
-  - "realm (0.2.1+alpha)":
+  - path_provider_ios (0.0.1):
+    - Flutter
+  - realm (0.3.0-beta):
```
desistefanova, line 5:
```suggestion
  - realm (0.3.0+beta):
```

flutter/macos/Podfile.lock
```diff
@@ -1,6 +1,6 @@
 PODS:
   - FlutterMacOS (1.0.0)
-  - "realm (0.2.1+alpha)":
+  - realm (0.3.0-beta):
```
desistefanova, line 3:
```suggestion
  - realm (0.3.0+beta):
```

flutter/pubspec.yaml
```diff
@@ -35,9 +35,9 @@ dependencies:
   # The following adds the Cupertino Icons font to your application.
   # Use with the CupertinoIcons class for iOS style icons.
   cupertino_icons: ^1.0.2
-  realm:
-    path: ./realm_packages/realm/
+  realm: 0.3.0-beta
```
desistefanova, line 38:
```suggestion
  realm: 0.3.0+beta
```

desistefanova (commented):
0.3.0+beta is available

flutter/lib/components/todo_list.dart
```diff
@@ -4,16 +4,15 @@ import 'package:realm/realm.dart';
 import 'todo_item.dart';
 import 'package:flutter_todo/realm/schemas.dart';
 
-const _animationDuration = Duration(milliseconds: 300);
-
 class TodoViewModel {
   final String id;
```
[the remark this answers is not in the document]
mongodben: i'd rather hold off on these changes for the V1 of this app before MDBW, unless you can implement. i'm not too familiar w the MVVM pattern (it was actually @nielsenko who added the code that you're commenting on).

happy to work together to do a fast follow with these changes post-MDBW, when we have a bit more breathing room.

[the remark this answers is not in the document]
nielsenko: Problem is that due to the animations of deleted you need the state of the TODO object after it has been deleted. Once we support frozen objects the `TodoViewModel` wrapper can be avoided.

We can also do an ugly delete animation, that doesn't show the data of the `Todo` object, but I think it is good to illustrate the issue to our users.

cbush (approved):

PR #70, opened 2022-06-06
Flex sync permissions demos

cbush:


cbush:
Merging so the demos might be available for Monday morning testing.

PR #78, opened 2022-06-17
fix flutter sdk version

mongodben:
update flutter sdk version b/c of bug with the Realm SDK `0.3.0+beta` release that doesn't download the Realm binaries correctly. 

see https://github.com/realm/realm-dart/blob/master/CHANGELOG.md#031beta-2022-06-07


dacharyc (approved):
LGTM but looks like the branch needs to be updated.

PR #81, opened 2022-06-27
Update Flutter app to use ObjectId as PK

mongodben:
done to make more inline with standard practices and other template apps

relates to https://github.com/mongodb-university/realm-template-apps/pull/79


MongoCaleb (approved):
LGTM

PR #82, opened 2022-07-13
Kotlin SDK Listener and Coroutine Improvements

nathan-contino-mongo:
- Added a Flow/listener for granular updates to recycler view data
- Switched to async coroutines on a background thread instead of blocking UI thread reads/writes
- Tweaked colors for better dark mode compatibility
- Limited application to portrait orientation since the view doesn't look good in landscape and Android requires a lot of extra boilerplate to avoid losing state on orientation switches.


mongomoe (approved):
LGTM 👍  + I learned what a coroutine is 😃

PR #86, opened 2022-08-12
Update RN for Flex Sync tutorial

nlarew:


generated/react-native-flexible-sync/package.json
```diff
@@ -7,22 +7,27 @@
     "ios": "react-native run-ios",
     "start": "react-native start",
     "test": "jest",
-    "lint": "eslint ."
+    "lint": "eslint .",
+    "prettier": "prettier --write src/*",
+    "generate:prod-partition-based-sync": "npx bluehawk copy . --output=../generated/react-native-partition-based-sync --ignore=realm.json --state=partition-based-sync",
+    "generate:prod-flexible-sync": "npx bluehawk copy . --output=../generated/react-native-flexible-sync --ignore=realm.json --state=flexible-sync",
+    "generate:prod": "npm run generate:prod-partition-based-sync && npm run generate:prod-flexible-sync"
```
mongomoe, line 14: Good idea to have these bluehawk commands in the package.json, but I suggest we shorten the names:
- generate-both # since it generates both FS and PBS; or alternatively 'generate-all' or just 'generate'
- generate-pbs
- generate-fs

react-native/package.json
```diff
@@ -8,22 +8,26 @@
     "start": "react-native start",
     "test": "jest",
     "lint": "eslint .",
-    "prettier": "prettier --write src/*"
+    "prettier": "prettier --write src/*",
+    "generate:prod-partition-based-sync": "npx bluehawk copy . --output=../generated/react-native-partition-based-sync --ignore=realm.json --state=partition-based-sync",
+    "generate:prod-flexible-sync": "npx bluehawk copy . --output=../generated/react-native-flexible-sync --ignore=realm.json --state=flexible-sync",
+    "generate:prod": "npm run generate:prod-partition-based-sync && npm run generate:prod-flexible-sync"
```
mongomoe, line 14: Good idea to have these bluehawk commands in the package.json, but I suggest we shorten the names:
- generate-both # since it generates both FS and PBS; or alternatively 'generate-all' or just 'generate'
- generate-pbs
- generate-fs

react-native/src/RealmContext.js
```diff
@@ -0,0 +1,6 @@
+import {createRealmContext} from '@realm/react';
+import { Task } from './TaskSchema';
+
+export default createRealmContext({
+  schema: [Task.schema],
+});
```
mongomoe, line 6: Is the idea behind separating the createRealmContext & schema so that they can be presented as 2 different steps in the tutorial?

If not, it might be worth keeping them combined as one file since it functionally does the same thing, and is a bit simpler. I think it gives the template-app-user one single blackbox (until they want to learn more) rather than two blackboxes. But that's just a suggestion and if you prefer it this way, feel free to keep it this way.

react-native/src/TasksView.js
```diff
@@ -18,37 +17,37 @@ export function TasksView() {
...
   //   // initialize the subscriptions
   //   const initSubscription = async () => {
   //     await realm.subscriptions.update(mutableSubs => {
-  //       mutableSubs.add(
-  //         realm.objects('Task').filtered(`owner_id == "${user.id}"`),
-  //       ); // subscribe to all Tasks of the logged in user
+  //       // subscribe to all Tasks of the logged in user
+  //       const ownTasks = realm
+  //         .objects('Task')
+  //         .filtered(`owner_id == "${user.id}"`);
```
mongomoe, line 30: Is this what we decided on adding to the subs? Didn't we say we wanted to add a subscription to priority or is that not intended to be in this & will only be shown on the tutorial doc?

mongomoe (approved):
LGTM, awesome job

react-native/src/RealmContext.js
nlarew: Yeah it will be a new step in the tut but I think this makes sense in the template as well. Co-locating with `TaskSchema` is simpler for just one schema but once you add a second schema then you'd want to refactor to something like this. Given that a lot of apps will use multiple schemas I think we should just give them the pattern up front.

react-native/package.json
nlarew: I'm using the same convention here as we have in the [`web-js`](https://github.com/mongodb-university/realm-template-apps/blob/main/web-js/package.json#L28) template but I agree we could actually be a little simpler here. I don't want to use the shorter names because these map to the Bluehawk state names to keep everything simpler.

Gonna go with

- `npm run generate` - generate all of this project's artifact repos
- `npm run generate:<state>` - generate a specific artifact repo

PR #88, opened 2022-08-15
(DOCSP-23969) Task -> Item + other tweaks for flex sync tutorial

nlarew:


generated/react-native-partition-based-sync/src/ItemSchema.js
```diff
@@ -0,0 +1,19 @@
+import {BSON} from 'realm';
+
+export class Item {
+  constructor({id = new BSON.ObjectId(), isComplete = false}) {
+    this.isComplete = isComplete;
+    this._id = id;
+  }
+
+  static schema = {
+    name: 'Task',
```
mongomoe, line 10:
```suggestion
    name: 'Item',
```

Looks like you forgot to change this code to 'Item'

react-native/src/ItemSchema.js
```diff
@@ -1,14 +1,18 @@
...
-export class Task {
+export class Item {
   constructor({id = new BSON.ObjectId(), isComplete = false}) {
     this.isComplete = isComplete;
     this._id = id;
   }
 
   static schema = {
+    // :state-start: partition-based-sync
     name: 'Task',
```
mongomoe, line 11: Looks like you forgot to change this code to 'Item' for the partition-based-sync state

mongomoe, line 11:
```suggestion
    name: 'Item',
```
Looks like you forgot to change this code to 'Item' for the partition-based-sync state

mongomoe:
The code is missing an initial subscription, can we add that. I feel it's an important part of the app that the other SDKs all have as well. You can see how to open a realm with an initial subscription for flexible sync here: https://www.mongodb.com/docs/realm/sdk/react-native/examples/flexible-sync/#bootstrap-the-realm-with-initial-subscriptions 

You can probably add it directly to your 'RealmProvider' tag in App.js:

```
const initialSubs =  {
    update: (subs, realm) => {
      subs.add(realm.objects("Team").filtered(`owner_id == "${user.id}'`) );
    },
};

return(
...
 <RealmProvider 
     sync={{flexible: true, initialSubscriptions: initialSubs}}
     fallback={ // ...}
    >
    ...
    </RealmProvider>
...
)
```

mongomoe:
Very close to LGTM, needs a couple fixes. Fixing the schema name in  `react-native/src/ItemSchema.js` is a mandatory fix since I get this error when I first start the app (without the fix):

error-object-item-not-found-in-schema

The comment about adding an initial subscription is a strong suggestion, but it's technically fine without it (I just think we should have it to match the other templates + I think its an important feature that stakeholders may want)

generated/react-native-flexible-sync/src/TaskSchema.js
```diff
@@ -1,5 +1,4 @@
 import {BSON} from 'realm';
```
mongomoe, line 1: I think you have both a 'TaskSchema.js' and a 'ItemSchema.js' file in both versions of the generated folders. Can you fix this?

react-native/src/ItemSchema.js
nlarew: This is actually intentional - the PBS backend still uses `Task` and it's hardcoded into the old backend

nlarew:
@mongomoe updated per your comments & rebased on top of the latest `phase1` commits

generated/react-native-flexible-sync/realm.json
```diff
@@ -1,4 +0,0 @@
-{
-  "appId": "todo-sync-zpqwm",
```
mongomoe, line 2: Needed fix: The generated files should have a `realm.json`.

generated/react-native-flexible-sync/src/RealmContext.js
```diff
@@ -0,0 +1,6 @@
+import {createRealmContext} from '@realm/react';
+import { Item } from './ItemSchema';
```
takameyer, line 2: Theres a linter error here.  This can be seen and fixed with:
```
npm run lint
npm run lint -- --fix
```

generated/react-native-flexible-sync/realm.json
nlarew: I don't think this is technically required since the "create a new template app" process that end users go through adds a `realm.json`. We ignore `realm.json` for web which is why it's ignored here too. That said, if it makes it easier to develop by including it here we can add it back.

nlarew:
@takameyer @mongomoe I updated the app to use Realm 10.20 + React 18 + RN 69.5

takameyer (changes requested):
remove `ios/app*` in `react-native` and `generated` projects.

generated/react-native-flexible-sync/realm.json
mongomoe: oh i see, it gave me an error when just cloning the source directly, but if it's created through that process its fine

mongomoe (approved):
LGTM 👍

takameyer (approved):

PR #100, opened 2022-10-11
add linux support

blagoev:
This adds Linux suport to the Flutter template app. 
Also it renames the `flutter` directory to `flutter_todo`. It is a bad practice to name the folder differently than the package inside and most importantly flutter tooling does not work if is not named like that.


mongodben (approved):
lgtm with a couple questions: 
1. is there anything in here for linux support that we ought to document?
2. what did you do to make this support linux (i see the generated files)?
3. how did you generate these files?

blagoev:
1) imo, we should not document anything more for Linux here
2) when you have an existing app you do `flutter create --platforms=linux` and this will add Linux support to your app which did not have it previously. This generates the `linux` directory in the app similar to `windows` and `mac`. This directory contains all build files necessary to build a Flutter Desktop Linux app
3) see 2) and more info here https://docs.flutter.dev/development/platform-integration/desktop#add-desktop-support-to-an-existing-flutter-app

blagoev:
@mongodben I will leave it to you to merge this

mongodben:
thx, will test out locally then merge

mongodben:
sorry for delay here. the code changes lgtm. i also updated the `manifest.json` file so that the template app generator is aware of the new directory name. 

@cbush could you take a quick look at this before we merge in?

PR #101, opened 2022-10-28
(DOCSP-26198): Comment on where to add in tutorial

mongodben:
added comments throughout the app for when users need to add code in the tutorial.

done based on feedback from Google Developer Advocate. 

I'm not 100% convinced this is the correct approach, but it was easy enough to implement and wanted to get team feedback here.

### DIFF
--- a/flutter_todo/README.md
+++ b/flutter_todo/README.md
@@ -1,6 +1,7 @@
 # Realm Flutter Todo
 
-A todo list application built with the [Realm Flutter SDK](https://www.mongodb.com/docs/realm/sdk/flutter/).
+A todo list application built with the [Realm Flutter SDK](https://www.mongodb.com/docs/realm/sdk/flutter/)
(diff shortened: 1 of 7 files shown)


mongodben:
closing per slack convo w @dacharyc. she said (which i fundamentally agree with):

>I think I don't love the comments. In theory, it sounds like a simple and non-invasive idea, but in practice, it feels like it's pushing me toward the tutorial pretty heavily - like some kind of marketing - and it also feels like a lot of stuff I'm going to want to remove if I'm not following the tutorial.
>
>I could potentially see following a tutorial once, but starting from a template app multiple times - every time I start a new app POC - and wouldn't want to have to keep dealing with those comments in the template.
>
> Also, we did talk about annotating the template app with comments explaining the "why" of some of the stuff. That was going to be our "explainer" piece for the clients themselves. I think these comments would interfere with that.

revisiting how to do the annotations in a separate PR. will add the README changes in a different PR.

PR #104, opened 2022-11-18
Add TypeScript Template

takameyer:
This creates a typescript version of the template application


react-native-ts/src/ItemListView.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+export function ItemListView() {
+  const realm = useRealm();
+  const items = useQuery(Item);
+  const user = useUser();
+  const [showNewItemOverlay, setShowNewItemOverlay] = useState(false);
+
+  useEffect(() => {
+    // initialize the subscriptions
+    const updateSubscriptions = async () => {
```
cbush, line 24: Does this effect need a cleanup or is it fire and forget? i.e. what happens if the ItemListView unmounts before realm.subscriptions.update() completion? If it's fine, a comment might help. Do subscriptions need to be cleaned up? (`mutableSubs.remove(...)`)

react-native-ts/src/ItemListView.tsx
```diff
@@ -0,0 +1,156 @@
...
+      });
+    };
+    updateSubscriptions();
+  }, [realm, user]);
+
+  // createItem() takes in a summary and then creates an Item object with that summary
+  const createItem = useCallback(
+    ({summary}: {summary: string}) => {
+      // if the realm exists, create an Item
+      if (realm && user) {
```
cbush, line 43: We're inconsistently checking the existence of realm. `useRealm()`'s signature assures us it will be a valid reference. If there's a case where it's not true, it should be updated in the SDK. If that's the case, maybe we should be a bit more strict than just silently ignoring the error state.

cbush: We also know that we have to be logged in for this component to be mounted, so I'm also wondering about useUser() and the checks for user. This might be a case for `assert(user)` since mounting this view without a user might be a developer error?

react-native-ts/src/LogoutButton.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react';
+import {Button, Alert} from 'react-native';
+import {useUser} from '@realm/react';
+
+export function LogoutButton() {
+  const user = useUser();
+  // The signOut function calls the logOut function on the currently
+  // logged in user and then navigates to the welcome screen
+  const signOut = () => {
```
cbush, line 9: No useCallback()?

cbush (approved):
Awesome, looks great, thank you! Tested on iOS, looks good. See comments.

react-native-ts/.prettierrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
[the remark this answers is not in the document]
takameyer: It comes from `react-native init`.  To upgrade React Native, I generally just generate a brand new project and copy the JavaScript files into it.

react-native-ts/src/ItemListView.tsx
takameyer: Good point, this should return a function that removes the subscription.  At some point I want to introduce a hook called `useSubscription` that automatically does this.

react-native-ts/.prettierrc.js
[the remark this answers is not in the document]
cbush: Perfect, thanks!

react-native-ts/src/ItemListView.tsx
takameyer: I've fixed the if statements.  I'll make an improvement to `@realm/react` later to fix the fact that user needs to be checked for `undefined` everywhere.  For now i'll just use the handy optional chaining syntax `user?.logout()`.

takameyer:
@cbush These changes are ready to be merged.  I'll do an upcoming PR for the functional changes.

PR #109, opened 2022-11-25
Add Features to the Template App

takameyer:
* Offline mode for trying out sync
* Completed items toggle that updates the flexible sync subscription
* Various improvements and fixes as found

PR #113, opened 2022-12-06
Add Flexible Sync Features to the Template App

ericjordanmossman:
* Add an offline mode for trying out sync.
* Add a toggle that updates the flexible sync subscription.
* Created a basic error handler.


ericjordanmossman:
@cbush I created a new v2 directory. Would you prefer I base it off the existing swiftui folder so it's a little cleaner like #109?

cbush:
v2 is perfect. 109 in fact goes into a new folder as well (before we started talking about "v2"), so all good!

v2/swiftui/App/AppConfig.swift
```diff
@@ -0,0 +1,25 @@
...
+import RealmSwift
+
+/// Store the Realm app details to use when instantiating the app and
+/// when using the `@AsyncOpen` property wrapper to open the realm.
+struct AppConfig {
+    var appId: String
+    var baseUrl: String
+}
+
+
```
cbush, line 11: [tiny] delete line

v2/swiftui/App/Views/ContentView.swift
```diff
@@ -0,0 +1,31 @@
...
+struct ContentView: View {
+    @ObservedObject var app: RealmSwift.App
+    @EnvironmentObject var errorHandler: ErrorHandler
+
+    var body: some View {
+        if let user = app.currentUser {
+            // Setup configuraton so user initially subscribes to their own tasks
+            let config = user.flexibleSyncConfiguration(initialSubscriptions: { subs in
+                subs.remove(named: "all_items")
+                if let _ = subs.first(named: "user_tasks") {
```
cbush, line 13: We might want to use a constant or enum to store these names. I also think we should keep the naming parallel (rather than "all ITEMS" -> "user TASKS". Plus, maybe "my" instead of "user")

v2/swiftui/App/Views/ItemList.swift
```diff
@@ -0,0 +1,25 @@
...
+    @ObservedResults(Item.self) var items
+    
+    var body: some View {
+        VStack {
+            List {
+                ForEach(items) { item in
+                    ItemRow(item: item)
+                }.onDelete(perform: $items.remove)
+            }
+            .listStyle(InsetListStyle())
```
cbush, line 18: Format: this line should probably be on the previous line (connected to `}`) or indented 1 more (compare .frame below)

v2/swiftui/App/Views/ItemsView.swift
```diff
@@ -0,0 +1,45 @@
...
+                    Toggle("Show Only My Tasks", isOn: $showMyItems).padding()
+                    ItemList()
+                }
+            }
+            .navigationBarItems(leading: self.leadingBarButton,
+                                trailing: HStack {
+                Button {
+                    syncEnabled = !syncEnabled
+                } label: {
+                    syncEnabled ? Image(systemName: "wifi") : Image(systemName: "wifi.slash")
```
cbush, line 35: FWIW, it was agreed on to call it "offline mode". If we reflect that in the code it might help sync beginners know what's going on a bit better.
```suggestion
                    !isInOfflineMode ? Image(systemName: "wifi") : Image(systemName: "wifi.slash")
```

v2/swiftui/App/Views/OpenRealmView.swift
```diff
@@ -0,0 +1,66 @@
...
+    // We must pass the user, so we can set the user.id when we create Item objects
+    @State var user: User
+    @State var showMyItems = true
+    @State var syncEnabled = true
+    // Configuration used to open the realm.
+    @Environment(\.realmConfiguration) private var config
+
+    var body: some View {
+        switch asyncOpen {
+            // Starting the Realm.asyncOpen process.
```
cbush, line 16: Very very minor but for some reason I think all of these comments that are above the `case` actually belong below the case.

cbush (approved):
Looks great. Thanks! A few very minor comments.

v2/swiftui/App/Views/ItemList.swift
ericjordanmossman: I'll move .onDelete a line down, so it's consistent with the rest of the attributes and with Apple's examples.

PR #118, opened 2022-12-10
Add more robust manifest validator

cbush:
...in preparation for more refactoring


tools/validate-manifest/src/Manifest.ts
```diff
@@ -0,0 +1,200 @@
...
+  A Manifest maps template IDs to template app specifications.
+ */
+export type Manifest = { [id: string]: TemplateSpec };
+
+/**
+  A template app is a backend configuration plus (optionally) some frontend
+  client source code.
+ */
+export type TemplateSpec = BackendOnlyTemplateSpec &
+  Partial<ClientTemplateSpec>;
```
mongodben, line 13: TIL i learned about TS `Partial` type

tools/validate-manifest/src/Manifest.ts
```diff
@@ -0,0 +1,200 @@
...
+  /**
+    The path to the exported App Services backend configuration directory.
+   */
+  backend_path: string;
+};
+
+/**
+  A template app can have a client.
+ */
+export type ClientTemplateSpec = BackendOnlyTemplateSpec & {
```
mongodben, line 43: nit: would it make sense to rename this here since it's a backend _and_ client template spec?

.github/workflows/check-manifest.yml
```diff
@@ -4,12 +4,16 @@ on: pull_request
...
-          node-version: 15.x
-      - run: npx ajv-cli validate -s manifest.schema.json -d manifest.json
+          node-version: 16.x
+      - name: Validate Manifest
+        working-directory: tools/validate-manifest
+        run: |
+          npm ci
+          npm run build
+          npm run test
+          node . validate ../../manifest.json
```
mongodben, line 19: i'm a little confused by the syntax here with `node . validate`.

could you make validate a script in the package.json file so the command is smthn like `npm run validate ../..manifest.json`?

mongodben (commented):
generally lgtm with a couple small nits/suggestions

---

also higher level question that is more for my own education (rather than any action to take here:

why did you structure the `validate-manifest` tool as a CLI, rather than a script to be run like `node myScript.js manifest.json` (or `tsc myScript.ts manifest.json`)?

mongodben:
> ts-node 

yep, that was the one i was thinking of

> What is the difference between a "CLI" and a "script to be run" in your understanding?

in my mind it's more qualitative. CLI more robust, multiple commands, vs 'script to be run' which is more one off.

cbush:
I have a template that I use to get these things started - https://github.com/cbush/typescript-cli-template - which also sets up prettier, tests, and all the other stuff I can't live without for development. In the end it is faster to make a more "robust" thing than a crappy one-off JS script.

tools/validate-manifest/src/Manifest.ts
cbush: My thinking was that the more important thing is that it defines a client and merely refers to a backend. I had `WithClientTemplateSpec` at one point but it seemed excessive.

.github/workflows/check-manifest.yml
cbush: Done, though I'm not sure it's actually that much clearer either way. Oh well!

mongodben (approved):

PR #127, opened 2023-01-11
Use version 10.32.3

ericjordanmossman:
- Use version realm 10.32.3 in the swiftui template app


dacharyc (approved):
Awesome - thanks for the quick fix, @ericjordanmossman !

PR #128, opened 2023-01-12
(DOCSP-26230): create Flutter template app tutorial version

mongodben:
Create Flutter template app tutorial version using Bluehawk state. 

> ❗ **Important** ❗ 
>
> If we proceed with merging this, let's do only after the initial V2 apps are merged to limit complexity.


.github/workflows/push-artifact-flutter.yml
```diff
@@ -5,15 +5,23 @@ on:
...
 jobs:
   push:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+
+      - name: copy template-realm.json to app pre-bluehawking
+        uses: canastro/copy-file-action@master
+        source: ./sync-todo/v2/template-realm.json
+        target: ./sync-todo/v2/client/flutter/assets/config/realm.json
```
cbush, line 19: Check how this is done in the bluehawk.sh instead on v1. More consistently if it's done as part of the bluehawk step (unfortunately bluehawk™ doesn't currently have a way to do this, and JSON doesn't allow comments)

cbush (commented):
Looks good. Why not add the code for the tutorial too? We could extract snippets that way. 🤔

.github/workflows/push-artifact-flutter.yml
mongodben: there was some reason that i did it here instead of the bluehawk step, which now escapes me. 

next week, will revisit this and see if i can figure out what my reasoning is now. if not,, will do similar to the v1 bluehawk.sh script

mongodben:
> Looks good. Why not add the code for the tutorial too? We could extract snippets that way. 🤔

it'd be too complicated to have the tutorial code changes going on in the same app. it'd be very hard to read and maintain. 

tho i do have the tutorial code bluehawked, just in a separate app (on another git branch), which you can see here - https://github.com/mongodb-university/realm-template-apps/tree/sync-todo-v2-final/sync-todo/v2/client/flutter

PR #130, opened 2023-01-16
Maui update - replace single edit alert with more extendable edit page.

MongoCaleb:


sync-todo/v2/client/maui/RealmTodo/ViewModels/ItemsViewModel.cs
```diff
@@ -47,16 +47,8 @@ public async Task Logout()
...
-
-            if (!string.IsNullOrEmpty(promptResult))
-            {
-                await realm.WriteAsync(() =>
-                {
-                    var item = new Item() { OwnerId = currentUserId, IsComplete = false, Summary = promptResult };
-                    realm.Add(item);
-                });
-            }
+            var item = new Dictionary<string, object>() { { "item", new Item() } };
```
papafe, line 50: This can't be called `item`, it's confusing. `navigationParameters` makes more sense to call it `itemParameter` like in the other method

sync-todo/v2/client/maui/RealmTodo/ViewModels/EditItemViewModel.cs
```diff
@@ -0,0 +1,57 @@
...
+    public partial class EditItemViewModel : BaseViewModel, IQueryAttributable
+    {
+        [ObservableProperty]
+        private Item newOrChangedItem;
+
+        [ObservableProperty]
+        private string dialogTitle;
+
+        private Realms.Realm realm;
+        private string currentUserId;
```
papafe, line 18: We are not using this anywhere else than "ApplyQueryAttributes", this can be removed and we can just use a local variable there or even use "RealmService.CurrentUser.Id" when needed

sync-todo/v2/client/maui/RealmTodo/ViewModels/EditItemViewModel.cs
```diff
@@ -0,0 +1,57 @@
...
+
+namespace RealmTodo.ViewModels
+{
+    public partial class EditItemViewModel : BaseViewModel, IQueryAttributable
+    {
+        [ObservableProperty]
+        private Item newOrChangedItem;
+
+        [ObservableProperty]
+        private string dialogTitle;
```
papafe, line 15: I think the name dialogTitle is a little confusing

papafe (changes requested):

sync-todo/v2/client/maui/RealmTodo/ViewModels/EditItemViewModel.cs
MongoCaleb: I did the same with "realm" because it is also only used once in the same method.

sync-todo/v2/client/maui/RealmTodo/ViewModels/ItemsViewModel.cs
```diff
@@ -66,16 +58,8 @@ public async Task EditItem(Item item)
...
-            var promptResult = await DialogService.ShowPromptAsync("Edit item", item.Summary);
-
-            if (!string.IsNullOrEmpty(promptResult))
-            {
-                await realm.WriteAsync(() =>
-                {
-                    item.Summary = promptResult;
-                });
-            }
+            var itemParameter = new Dictionary<string, object>() { { "item", item } };
```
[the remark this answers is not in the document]
MongoCaleb: As per our Slack convo, I've gone with option (B) because we want the user to be able to "cancel" their changes -- mimicking "drafts".

sync-todo/v2/client/maui/RealmTodo/AppShell.xaml
```diff
@@ -16,5 +16,10 @@
         ContentTemplate="{DataTemplate local:ItemsPage}"
         Route="items" />
 
+    <ShellContent
```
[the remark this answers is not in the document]
MongoCaleb: Out of curiosity, could we not also define it in the xaml as a nested route, like this?
```
<ShellContent
        Title="RealmTodo"
        ContentTemplate="{DataTemplate local:ItemsPage}"
        Route="items" >
        <ShellContent
            Title="RealmTodo"
            ContentTemplate="{DataTemplate local:EditItemPage}"
            Route="editItem" />
    </ShellContent>
```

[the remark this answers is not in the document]
papafe: In this case it could work, but it's a little different. If you define it in the xaml, then it means that if you show EditItemPage, then there must be ItemsPage just behind.
If you define it in code, you're creating a "global routes", that can be called from anywhere in the app. This is generally useful for "detail pages". For example, in a todo app you maybe want to be able to see/modify the todo from different parts of the application.

sync-todo/v2/client/maui/RealmTodo/Views/EditItemPage.xaml
```diff
@@ -0,0 +1,24 @@
...
+        <vm:EditItemViewModel />
+    </ContentPage.BindingContext>
+     <VerticalStackLayout Padding="10">
+        <Label 
+            Text="{Binding PageHeader}"
+            VerticalOptions="Center" 
+            HorizontalOptions="Center"
+            TextColor="{DynamicResource Primary}"/>
+        <Entry Text="{Binding Summary}" Placeholder="Item Summary"  PlaceholderColor="#787878"/>
+
```
papafe, line 17: Do we need the extra space?

papafe (approved):
Looks good! The only thing I would say to consider is that maybe we can remove the cancel button, given that there should be already a back button automatically that essentially does the same. 
Also, I didn't try to run the code, but I trust in your UI capabilities 😁

PR #131, opened 2023-01-17
Remove lockfiles and workspace files

takameyer:
These should be created by the user of the template with `npx pod-install`.  Otherwise, it may cause setup problems.


cbush:
Should they be added to the local .gitignore then?

takameyer:
Yes and no.  What would be ideal is a template `.gitignore`, which overrides our project `.gitignore` on template initialization.  The `Podfile.lock` should not be ignored when the template is downloaded from the user.  They should create it themselves when they initialize the project and commit it to their repo.  The react native templates have a mechanism for this.  Hidden files are prefixed with `_` and rewritten when the template is generated.
https://github.com/realm/realm-js/tree/master/templates/react-native-template-realm-ts/template

cbush:
Interesting. I think we will have the ability to do that since we should actually clone/copy/pull from the generated directory anyway, like this: https://github.com/mongodb-university/realm-template-apps/tree/main/sync-todo/v1

This allows us to remove the app ID from the realm.json file before anybody clones the app, but we can also use it to inject a special .gitignore

cbush (approved):

PR #132, opened 2023-01-17
V2 launch

cbush:
- Realm.json -> atlasConfig.json (etc)
- Set up bluehawk generated/ structure like in the v1 directory for the v2 apps (so that config json gets cleaned up)
- Update arty-fact so it pushes from the v2 apps instead of v1


manifest.json
```diff
@@ -65,16 +65,56 @@
     "metadata_filename": "realm",
     "file_format": "json"
   },
-  "react-native-ts.todo.flex": {
+  "react-native.todo.v2": {
```
dacharyc, line 68: Ah! I didn't realize we were updating the names of these. When we had talked about this before, we were going to leave the names the same so this would be a drop-in replacement for V1. 

Is there a reason we're making this a v2 instead of just replacing? They run against a different backend, so I think most people are just going to want "the client" that matches the Device Sync app they create in the UI. Having to think about `react-native-todo.flex` vs `react-native.todo.v2` to decide which one to get (if they do it via CLI) is maybe an unnecessary point of friction?

sync-todo/v2/client/react-native/README.md
```diff
@@ -1,17 +1,20 @@
...
 - Set up an application in App Services.
-    -  You can use the [`realm-cli`](https://www.mongodb.com/docs/atlas/app-services/cli/) to set up the backend and download a copy of this template:
-		```
-		realm-cli apps create -n "<App Name>" --template "react-native-ts.todo.flex"
-		```
+  - You can use the [`realm-cli`](https://www.mongodb.com/docs/atlas/app-services/cli/) to set up the backend and download a copy of this template:
+
+
+      ```
+      realm-cli apps create -n "<App Name>" --template "react-native-ts.todo.flex"
```
dacharyc, line 12: This README references the client ID so we should update this.
```suggestion
      realm-cli apps create -n "<App Name>" --template "react-native.todo.v2"
```

sync-todo/v2/client/swiftui/App/AppConfig.swift
```diff
@@ -8,12 +8,12 @@ struct AppConfig {
     var baseUrl: String
 }
 
-/// Read the Realm.plist file and store the app ID and baseUrl to use elsewhere.
+/// Read the atlasConfig.plist file and store the app ID and baseUrl to use elsewhere.
 func loadAppConfig() -> AppConfig {
     guard let path = Bundle.main.path(forResource: "Realm", ofType: "plist") else {
```
dacharyc, line 13:
```suggestion
    guard let path = Bundle.main.path(forResource: "atlasConfig", ofType: "plist") else {
```

sync-todo/v2/client/swiftui/App/AppConfig.swift
```diff
@@ -8,12 +8,12 @@ struct AppConfig {
...
-        fatalError("Could not load Realm.plist file!")
+        fatalError("Could not load atlasConfig.plist file!")
     }
-    // Any errors here indicate that the Realm.plist file has not been formatted properly.
+    // Any errors here indicate that the atlasConfig.plist file has not been formatted properly.
     // Expected key/values:
     //      "appId": "your Realm app ID"
```
dacharyc, line 18:
```suggestion
    //      "appId": "your App Services app ID"
```

dacharyc: I can't comment directly on the lines below, but we should probably rename `realmPropertyList` in ln 20 below to `atlasConfigPropertyList` or something along those lines to go along with the renaming. And then update ln 21 & ln 22 to get the appId and baseUrl from it the renamed var.

sync-todo/v2/client/swiftui/App/realmApp.swift
```diff
@@ -1,7 +1,7 @@
 import SwiftUI
 import RealmSwift
 
-/// This method loads app config details from a Realm.plist we generate
+/// This method loads app config details from a atlasConfig.plist we generate
 /// for the template apps.
 /// When you create your own Realm app, use your preferred method
```
dacharyc, line 6: Maybe this?
```suggestion
/// When you create your own Atlas Device Sync app, use your preferred method
```

dacharyc: I can't comment directly on the line below, but should `realmApp` in ln 10 be renamed & places that use it be refactored to match the new name? (i.e. in this file ln 14, ln 18 - not sure if it's used elsewhere without doing some searching?)

dacharyc (changes requested):
Overall, this looks good. I spotted a couple of additional places that should definitely be updated, and had a couple of Qs. But we should update and regenerate the generated templates.

manifest.json
cbush: No, per the release plan, this is a temporary ID so we can test on the CLI without disrupting v1 - we will update it to be a drop-in replacement later.

sync-todo/v2/client/react-native/README.md
```diff
@@ -1,17 +1,20 @@
...
 - Set up an application in App Services.
-    -  You can use the [`realm-cli`](https://www.mongodb.com/docs/atlas/app-services/cli/) to set up the backend and download a copy of this template:
-		```
-		realm-cli apps create -n "<App Name>" --template "react-native-ts.todo.flex"
-		```
+  - You can use the [`realm-cli`](https://www.mongodb.com/docs/atlas/app-services/cli/) to set up the backend and download a copy of this template:
+
+
+      ```
+      realm-cli apps create -n "<App Name>" --template "react-native.todo.v2"
```
dacharyc, line 12: My suggestion about changing this name was with the understanding that we were going to a v2 name for these apps. Since it's temp, we probably shouldn't have changed it. But I don't think this needs to block merging the updates - just something to address when we finalize the name at the "real" launch.

dacharyc (approved):
Nice - LGTM, let's get testing!

PR #135, opened 2023-01-19
Update manifest: v2 -> main templates

cbush:


MongoCaleb (approved):
LGTM. I wonder about changing "flex", though, since we're no longer doing PBS templates and "flex" has a meaning of it's own that might confuse.

cbush:
I agree, but we're using the same IDs so that they are drop-in replacements on the UI. We can look into changing it later.

PR #138, opened 2023-02-08
flutter template app to v1.0.0

mongodben:


dacharyc (approved):
There are a lot of dependency updates here. Just wanted to make sure those were all intentional as part of updating the SDK version, and not accidental changes that have leaked in. Assuming it's intentional, LGTM 👍

PR #153, opened 2023-05-23
(DOCSP-30014) Update Flutter SDK template app versions

krollins-mdb:
Pending the next Flutter SDK release: update the Flutter template app versions in `pubspec.yaml`.

https://github.com/mongodb-university/realm-template-apps/blob/e359b675791d11c3f55d29e4e25bcba3964e857a/sync-todo/v2/client/flutter/pubspec.yaml

### DIFF
--- a/sync-todo/v2/client/flutter/pubspec.yaml
+++ b/sync-todo/v2/client/flutter/pubspec.yaml
@@ -6,7 +6,7 @@ publish_to: "none" # Remove this line if you wish to publish to pub.dev
 version: 1.0.0+1
 
 environment:
-  sdk: ">=2.17.0"
+  sdk: ">=2.17.5 <4.0.0"
 
 dependencies:
   flutter:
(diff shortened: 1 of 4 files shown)


sync-todo/v2/client/flutter/pubspec.yaml
```diff
@@ -6,15 +6,15 @@ publish_to: "none" # Remove this line if you wish to publish to pub.dev
 version: 1.0.0+1
 
 environment:
-  sdk: ">=2.17.0"
+  sdk: ">=2.17.5 <4.0.0"
```
dacharyc, line 9: Just a Q to clarify for me - I'm not familiar with this syntax but it looks like we're limiting the SDK here to versions less than 4.0.0, but we haven't done that previously. Just wanted to make sure this is intended. (Same with dart in the pubspec.lock)

dacharyc (approved):
Overall LGTM! Just a Q about the versions. Assuming the client runs fine locally after you update it, seems good to me.

sync-todo/v2/client/flutter/pubspec.yaml
krollins-mdb: Yep! Probably unnecessary, but it's what the Dart SDK team uses and suggested.

PR #155, opened 2023-05-25
add a sort to swift template app

sudssm:
This matches the behavior of other template apps, and ensures a consistent ordering in the view


sudssm:
@nlarew

dacharyc (changes requested):
Awesome - thank you for making this PR @sudssm ! 

One more thing we need - can you run `bluehawk.sh` in the `sync-todo/v2` directory and commit the result? This creates the `generated` client files we use to update the artifact repo and the App Services backend.

dacharyc (approved):
Awesome - thank you!

PR #175, opened 2023-11-15
BAAS-26383: Move rules, schema, relationships in JS template to todo/Item

malexandert:
JIRA: https://jira.mongodb.org/browse/BAAS-26383

(hopefully) the last template update 😅. The old parsing code would decide which database/collection a rule belonged to based on the `database` and `collection` fields in the json itself, but the current parsing code uses the file name. This was causing the rule to get parsed for the `todo.Task` namespace, instead of the `todo.Item` namespace

Again, this should be a no-op for the current production templates, but will be important going forward


dacharyc (approved):
Awesome - thank you!

PR #177, opened 2023-11-22
Update React Native version

### LINKED ISSUE #173: Realm Prop load error
When running the code through the android simulator, it errors out when initializing the RealmProvider. I get the following errors:

 BUNDLE  ./index.js

 LOG  View your data in MongoDB Atlas: https://cloud.mongodb.com/links/<...>/explorer/Cluster0/database/collection/find.
 ERROR  Error: Exception in HostObject::get for prop 'Realm': java.lang.UnsatisfiedLinkError: couldn't find DSO to load: librealm.so caused by: dlopen failed: cannot locate symbol "__emutls_get_address" referenced by "/data/app/~~UrHcuOA1X7s9x-M8jCfU4g==/com.synctutorial-q5oSJsHgqnnBeIlJ_Vt4TQ==/lib/x86_64/librealm.so"... result: 0, js engine: hermes
 ERROR  Invariant Violation: Failed to call into JavaScript module method AppRegistry.runApplication(). Module has not been registered as callable. Registered callable JavaScript modules (n = 10): Systrace, JSTimers, HeapCapture, SamplingProfiler, RCTLog, RCTDeviceEventEmitter, RCTNativeAppEventEmitter, GlobalPerformanceLogger, JSDevSupportModule, HMRClient.        A frequent cause of the error is that the application entry file path is incorrect. This can also happen when the JS bundle is corrupt or there is an early initialization error when loading React Native., js engine: hermes
 ERROR  Invariant Violation: Failed to call into JavaScript module method AppRegistry.runApplication(). Module has not been registered as callable. Registered callable JavaScript modules (n = 10): Systrace, JSTimers, HeapCapture, SamplingProfiler, RCTLog, RCTDeviceEventEmitter, RCTNativeAppEventEmitter, GlobalPerformanceLogger, JSDevSupportModule, HMRClient.        A frequent cause of the error is that the application entry file path is incorrect. This can also happen when the JS bundle is corrupt or there is an early initialization error when loading React Native., js engine: hermes

krollins-mdb:
Closes #173 

We have some React Native and package dependency weirdness in the template app. In this PR, I created a new React Native app, then migrated the old template app's code to it. This is much easier than trying to navigate upgrading the React Native version.

One big thing that didn't make the transition: icons. They were broken in the old app anyway. It's certainly possible to fix that, but in the interest of time, I wanted to get this updated app published. We can look at adding icons back in later.


sync-todo/v2/client/react-native/source/App.tsx
dacharyc, line 1: It looks like these files have moved from `src` to `source` in this PR. 

The [Device Sync RN Tutorial](https://www.mongodb.com/docs/atlas/app-services/tutorial/react-native/#react-native-app) instructs users to look for these files in `src`, so we probably need a PR to update that to go along with these changes.

(In looking at that tutorial more, it also instructs users to edit files in `src/filename` in each step in the "Add a Priority Level Field" procedure, so I guess those should be updated, too. Not sure if there are other changes that should be made?)

sync-todo/v2/client/react-native/src/RealmContext.ts
dacharyc, line 1: We have a line item for `RealmContext.ts` in the Device Sync RN tutorial, so a PR that updates that document should also update this. (I guess this is now AppWrapper and works a little differently?)

dacharyc (approved):
Overall LGTM. There are a couple of changes here that will affect the Device Sync RN tutorial, which I've called out in the comments. Those will require a separate ticket to update that documentation when we merge this. I do notice a couple of new files being committed (`Podfile.lock`, `SyncTutorial.xcworkspace/contents.xcworkspacedata`, `jest.config.js`). I just want to confirm that adding those files is intentional, and that we shouldn't be `.gitignore`-ing any of them.

krollins-mdb:
I have a PR up for the tutorial: https://github.com/mongodb/docs-app-services/pull/637

PR #187, opened 2024-02-09
DOCSP-35656: Update README for App Services backend

cbullinger:
Update README with steps outlining how to create an App Services App that uses the project's pre-configured backend. 

(These steps can only be tested once the `edge-server-template` branch is merged to `main` and the new `edge-server.todo` template is available via CLI.)

### DIFF
--- a/other/edge-server/client/README.md
+++ b/other/edge-server/client/README.md
@@ -77,7 +77,7 @@ directory in your current path containing the backend files:
 
 ```shell 
 appservices app create \
---name  EdgeServerApp \
+--name EdgeServerApp \
 --template edge-server.todo
 ```
 
--- a/other/edge-server/client/README.md
+++ b/other/edge-server/client/README.md
@@ -1,38 +1,116 @@
 # Edge Server with MongoDB Node.js Driver
 
-This example application uses the MongoDB Node.js Driver to connect to 
-Edge Server. It has a React frontend for performing CRUD operations. It uses 
-the MongoDB Node.js Driver in an Express server application to connect to
-the Edge Server.
+This example todo application demonstrates using Edge Server with a MongoDB Driver. It uses the MongoDB Node.js Driver in an Express server application to connect to 
+Edge Server. It has a React frontend for performing CRUD operations. 
 
 React client -> Express server (Node.js Driver) -> Edge Server -> Atlas
 
 To use this example application, you must:
 
-1. Create an App Services App using the backend configuration provided in
-  this project's [TODO: Cory insert directory name here] directory.
+1. Create an App Services App that uses the configured backend provided in 
+  this project's ../backend directory.
 2. Download, configure, and start the Edge Server.
 3. Install dependencies and start the Node.js Express server.
 4. Install dependencies and start the React server.
 
-This is a MERN stack that connects to the Edge Server instead of Atlas. For
-a MERN stack tutorial, refer to 
+This is a MERN stack that connects to the Edge Server instead of Atlas. 
+(Note that MongoDB wire protocol connections connect directly to Edge Server 
+without authorization.) 
+
+For a MERN stack tutorial, refer to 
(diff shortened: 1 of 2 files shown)


other/edge-server/client/README.md
```diff
@@ -1,38 +1,116 @@
...
+Edge Server. It has a React frontend for performing CRUD operations. 
 
 React client -> Express server (Node.js Driver) -> Edge Server -> Atlas
 
 To use this example application, you must:
 
-1. Create an App Services App using the backend configuration provided in
-  this project's [TODO: Cory insert directory name here] directory.
+1. Create an App Services App that uses the configured backend provided in 
+  this project's ../backend directory.
```
dacharyc, line 11: Since we're now using a template app for this, we won't include the backend directory in the artifact repo - so it doesn't make sense to refer to it here. 
```suggestion
1. Create an App Services App based on the `edge-server.todo` template app. Get Edge Server enabled for your new app.
```

other/edge-server/client/README.md
```diff
@@ -1,38 +1,116 @@
...
 2. Download, configure, and start the Edge Server.
 3. Install dependencies and start the Node.js Express server.
 4. Install dependencies and start the React server.
 
-This is a MERN stack that connects to the Edge Server instead of Atlas. For
-a MERN stack tutorial, refer to 
+This is a MERN stack that connects to the Edge Server instead of Atlas. 
+(Note that MongoDB wire protocol connections connect directly to Edge Server 
+without authorization.)
```
dacharyc, line 18: Adding authorization for wireprotocol connections is part of the Public Preview, so we should probably be more explicit here. We can update this app once authentication for wireprotocol connections is supported by Edge Server.

```suggestion
In this version of the example app, MongoDB wire protocol connections connect directly to Edge Server without authorization. 
```

other/edge-server/client/README.md
```diff
@@ -1,38 +1,116 @@
...
+For this example application, you'll create an App Services App based on the 
+`edge-server.todo` template app. 
+
+This template app uses a backend with the following enabled: 
+
+- Device Sync configured with:
+  - An already defined `Item` collection in a `todo` database
+  - A single default rule that allows *any* user to read or write Items to 
+    the collection
+- Anonymous authentication only
```
dacharyc, line 35: Hmm. I wonder if we can disable anonymous authentication and still use Edge Server? We might want to test that, and remove this line if it is possible. 

Mainly, I don't want to mislead people into thinking anonymous auth is important here somehow if it's not required. The wireprotocol connection doesn't use it, so I think if we do want to call out that anonymous authentication is configured - i.e. if it _is_ required - we have to say why.

(I _think_ it might have been enabled before because we were trying to make this interoperable with Device SDK clients, but we've found that with the required rules, we can't do that. So it might make sense to just turn it off.)

dacharyc: Also, I wonder if we should explicitly call out that this differs from the Device SDK todo backend, and that this app won't be interoperable with Device SDK template apps.

other/edge-server/client/README.md
```diff
@@ -1,38 +1,116 @@
...
+The following creates a new app named "EdgeServerApp" that uses the 
+template's pre-configured backend: 
 
-- What things are enabled in this App backend (Sync, anonymous auth, only a 
-  default rule that is open to the internet)
-- Maybe how this differs from the sync-todo backend? I don't know if we want
-  to contrast it with using the Device SDK todo template app as it's slightly different
+```shell 
+appservices app create \
+--name  EdgeServerApp \
```
dacharyc, line 75: Nit: extra space.
```suggestion
--name EdgeServerApp \
```

dacharyc (approved):
Couple of Qs and suggestions, including one that may require testing (seeing if we can do this without anonymous auth enabled), and if so, a small update to the backend. But overall LGTM. Thank you for working on this! 🙏

PR #190, opened 2024-02-13
Edge Server README fix 

cbullinger:


dacharyc (approved):
Thanks for catching these!

PR #192, opened 2024-02-16
DOCSP-36923: Bump Kotlin app to v1.13.0

cbullinger:
Update Kotlin template app from v1.8.0 to v1.13.0 and update any dependencies


dacharyc (approved):
Thank you for updating this! But we also need to run `bluehawk.sh` to copy these updates to the `generated` directory: https://github.com/mongodb-university/realm-template-apps/blob/main/sync-todo/v2/bluehawk.sh

After you do that, this is LGTM!

PR #200, opened 2024-04-09
Improve Edge Server template app DX

krollins-mdb:
This PR makes the following changes to the Edge Server template app:
- Added a quick start option. After setting up an app services app and setting up Edge Server, at the project root, users can run `npm run install` to install the Node server and React client dependencies. Then they can run `npm run start` to start the Edge Server, Node server, and React client. `npm run shut-down` shuts down all three pieces.
- Changed the Express server port from 5000 to 5055. Also updated the React client to refer to 5055. This is to avoid AirPlay on MacOS, which sometimes takes over port 5000.
- Hard-code an owner_id of {{"edge"}} for the Express client.

Notable not doing:
- Abstract port value so that the Express server and React client share the same value. This became a harder problem to solve than expected. If users need to change the port value, they need to change it in the Express server and React client code. This is already documented.


other/edge-server/README.md
```diff
@@ -286,14 +309,14 @@ This indicates that another device on your machine is using the port that
 the Express server is trying to use to listen for incoming connections.
 This example app uses port `5000` by default.
```
dacharyc, line 310:
```suggestion
This example app uses port `5055` by default.
```

other/edge-server/README.md
```diff
@@ -242,15 +265,15 @@ Install the required dependencies:
 npm install
 ```
 
-Add a `.env` file inside the `node-server` directory with the details 
+Add a `.env` file inside the `node-server` directory with the details
```
dacharyc, line 268: Oof. I am realizing now this is even more of a PITA since you've made it generally nicer about not having to install dependencies and start the servers individually.

At the very least, since this README is now above the `client` directory and the `node-server` directory isn't visible, we should probably add that detail.

But it would really be nicer to have some higher-level thing people can config instead of having to dig deeper to configure stuff.
```suggestion
Add a `.env` file inside the `client/node-server` directory with the details
```

(We should probably also remove the lines above about cding into `node-server` and installing dependencies, since you've now got that higher-level script that manages installing all the dependencies.)

other/edge-server/README.md
```diff
@@ -242,15 +265,15 @@ Install the required dependencies:
...
 ```env
 # This URI string is for Atlas Edge Server
 EDGE_SERVER_URI="mongodb://localhost:27021"
 
 # This port is for the Node.js server
-PORT=5000
+PORT=5055
 ```
 
 Then, run the Express server:
```
dacharyc, line 279: We can probably remove these lines about running the Express Server since that is done now at the higher level.

dacharyc: Can't comment below, but we should update the path in ln 269/292 to show that `node-server` is in the `client/` directory, and in ln 273/296 to say instead "at the root of your `client/node-server` directory.

other/edge-server/README.md
```diff
@@ -286,14 +309,14 @@ This indicates that another device on your machine is using the port that
 the Express server is trying to use to listen for incoming connections.
 This example app uses port `5000` by default.
 
-You can change the port used in your `.env` file at the root of the 
+You can change the port used in your `.env` file at the root of the
 `node-server` directory. If you change this port, you must also change
```
dacharyc, line 313:
```suggestion
`client/node-server` directory. If you change this port, you must also change
```

other/edge-server/README.md
```diff
@@ -286,14 +309,14 @@ This indicates that another device on your machine is using the port that
...
 the port that the React client uses to communicate with the Express server.
 Change the port in `react-client/src/endpoints.ts` on line 4 to match
```
dacharyc, line 315:
```suggestion
Change the port in `client/react-client/src/endpoints.ts` on line 4 to match
```

other/edge-server/README.md
```diff
@@ -306,13 +329,13 @@ connect ECONNREFUSED ::1:27021, connect ECONNREFUSED 127.0.0.1:27021
...
-that the Edge Server is running. If you have changed the URI and/or port where 
-the Edge Server is listening for wireprotocol connections, change the URI 
+that the Edge Server is running. If you have changed the URI and/or port where
+the Edge Server is listening for wireprotocol connections, change the URI
 and/or port in your `.env` file.
 
 ## Install dependencies and start the React server
 
-With the Edge Server and the Express server running, cd into the `react-client` 
+With the Edge Server and the Express server running, cd into the `react-client`
```
dacharyc, line 338: We can probably just delete this section about installing dependencies and starting the React server, since that can now be managed at the higher level with your new package.json.

(Also, we should document that as the way to install the dependencies and run the servers.)

other/edge-server/README.md
```diff
@@ -394,21 +417,24 @@ check:
...
 
-For information about how to check the App Services logs, refer to 
+For information about how to check the App Services logs, refer to
 [View Application Logs](https://www.mongodb.com/docs/atlas/app-services/activity/view-logs/)
 
 ## Issues
 
 Please report issues with the template at: https://github.com/mongodb-university/realm-template-apps/issues/new
+
+## Project npm commands
```
dacharyc, line 435: I would move this section up and remove the stuff about installing dependencies for the node server and the React client separately. This is the happy path and we want to present it front-and-center.

other/edge-server/README.md
```diff
@@ -372,7 +395,7 @@ the React client. The client does not have a listener observing changes
 to the collection. You can manually reload the React client to view the
 new `Item`.
 
-If you reload the React client and it still doesn't show the new `Item`, 
+If you reload the React client and it still doesn't show the new `Item`,
 check:
 
 - The Edge Server is running
```
dacharyc, line 401: We've added back the `owner_id` here, but it's not reflected in the schema below. We should update this schema to match the one we're now using in the backend.

other/edge-server/README.md
```diff
@@ -1,72 +1,86 @@
...
+This template app uses a backend with Atlas Device Sync enabled and configured with:
 
 - An already defined `Item` collection in a `todo` database
-- A single default rule that allows *any* user to read or write Items to 
+- A single default rule that allows _any_ user to read or write Items to
   the collection
 
-> **NOTE:** This example application is NOT interoperable with any of the Atlas 
-> Device SDK template apps. You must use the `edge-server.todo` template app. 
+> **NOTE:** This example application is NOT interoperable with any of the Atlas
```
dacharyc, line 58: I _think_ we can take this note out now. By adding the `owner_id` back, we should have made it interoperable with the Device SDK template apps. (We should also test this backend with the Device SDK template apps.)

You must still use `edge-server.todo` to _create_ a template app that the Edge Server template can connect to, but this should now work with the Device SDK templates.

dacharyc (commented):
Back to you with some comments re: the README.

Also, I'm curious - have you tested this with a Device SDK template app? It should now be interoperable, but we'll want to confirm that before we merge this.

dacharyc (commented):
Ah! Just realized our workflow also only pushes files that are _in_ the `client` directory to the artifact repo: https://github.com/mongodb-university/realm-template-apps/blob/main/.github/workflows/push-artifact-edge-server.yml

And the artifact repo looks like this: https://github.com/mongodb/template-app-edge-server

So if you want to have the README.md, the package-lock.json and the package.json all go up to the artifact repo, you'll need to edit the workflow. And I assume you'll also want to exclude the `backend` directory since we don't necessarily want that in the artifact repo.

This may be a good case for creating a `generated` directory and using Bluehawk copy to only copy over the files we want, excluding the `backend`, and push _that_ up to the artifact repo. I'll let you decide how you want to handle that.

other/edge-server/README.md
krollins-mdb: Yeah, would be nice to streamline this more. I'll do some more thinking about a good approach in the future.

other/edge-server/README.md
krollins-mdb: I reframed this section and the Express section as alternatives to the quick start and I link to the quick start above. The info in these sections still works if people want to do it on their own.

krollins-mdb: Just kidding. I've merged the Express and React sections and made it more of the happy path.

krollins-mdb:
Excellent point, Dachary! I don't know why I had it in my head that I needed this stuff at the `edge-server` level. I moved everything to the `client` directory and tested again. Should be good now.

dacharyc (approved):
Looking good! Just a couple of other small suggestions to clarify some of the backend stuff, but that's not really a blocker.

PR #202, opened 2024-04-26
(DOCSP-38891) Refactor to support email/password auth

krollins-mdb:
This PR refactors existing code to enable an email/password auth flow. The new structure is set up to easily add more auth types in the future.
- Refactor to support email/password auth
  - Move Edge Server connection logic to endpoint routes in Express server
  - Add new auth endpoints in Express server
  - Add a log in page to React client that will show until one of the auth endpoints responds with successful Edge Server connection
- Rename /node-server/ to /express-server/
- Update README to reflect new `edgectl` use

### Added after initial reviews:
**React Client**
- Persist user information in browser local storage
- Make the call for Edge Server Todo items repeat every 5 seconds to refresh data. This is a less-than-ideal way to implement the the refresh, but it works. Alternatives include adding a refresh button instead or building out a proper stream. The latter is more work than we have time for.

**Express server**
- Add proper REST error codes when API requests fail.

**Both**
- Add better error handling and messaging for API calls. Show notifications when calls fail.
- Generally clean up the code to use consistent standards


other/edge-server/backend/data_sources/mongodb-atlas/todo/Item/rules.json
```diff
@@ -6,7 +6,7 @@
             "name": "readAndWriteAll",
             "apply_when": {
                 "%%user.type": {
```
dacharyc, line 8: Engineering pointed out in one of my docs PRs that we can replace this expression with something simpler:

`"%%user.type": "edge"`

other/edge-server/backend/data_sources/mongodb-atlas/todo/Item/rules.json
```diff
@@ -6,7 +6,7 @@
             "name": "readAndWriteAll",
```
dacharyc, line 6: I suggest renaming this so we make it clearer this is the role that applies to the Edge Server. I know people can infer it from the `apply_when` expression, but it doesn't hurt to make it more explicit.

```suggestion
            "name": "edgeServerReadWriteAll",
```

other/edge-server/client/.gitignore
```diff
@@ -23,3 +23,5 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+docker-compose.yml
```
dacharyc, line 27: I don't think we need this in the .gitignore, because with the new `edgectl` stuff, this file will get written somewhere outside of this directory.

other/edge-server/client/README.md
```diff
@@ -34,32 +35,34 @@ previewing Edge Server, refer to the product page at
...
 
-   # This port is for the Node.js server
-   PORT=5055
+    # This port is for the Express server
+    PORT=5055
    ```
 
-6. Then run `npm run start`.
-7. Explore how CRUD works from both the React Client and the App Services UI.
+7. Then run `npm run start`.
```
dacharyc, line 59: I know this should go without saying, but more than once, I have cd'd into the `node-server` (now `express-server`) dir to add and populate the `.env` file, and forgotten to go back to the root directory to run `npm run start`.

```suggestion
7. From the project's root directory, run `npm run start`.
```

other/edge-server/client/README.md
```diff
@@ -9,15 +9,16 @@ React client -> Express server (Node.js Driver) -> Edge Server -> Atlas
 
 To use this example application, you must:
 
-1. Create an App Services App based on the `edge-server.todo` template app, and
-   get Edge Server enabled for your new app.
-2. Download, configure, and start the Edge Server.
-3. Install dependencies and start the Node.js Express server.
-4. Install dependencies and start the React server.
+1. Create an App Services App based on the `edge-server.todo` template app.
+2. Install Edge Server dependencies
```
dacharyc, line 13: The install script handles prompting users to install missing dependencies (I think) so I think we can remove this step from the README.

other/edge-server/client/README.md
dacharyc, line 1: Broadly, we should probably add _something_ to the README about how disabling auth is required to use the anonymous auth type. Maybe a step in the quick start? (i.e. the `edgectl config --insecure-disable-auth=true`)

I also wonder if we should mention that if you're running more than one Edge Server instance on the host hardware, you can't use the `npm run start` command and must instead start all of the servers separately. (Because you'll need to start the Edge Server with a `--profile` flag.)

dacharyc (commented):
Nice start! I've got a few suggestions for some minor tweaks, but I think overall it's working well and it's nice to have this ready to go - good job!

other/edge-server/client/README.md
sudssm: disabling auth is not actually required to use anonymous auth. if you have anonymous auth provider enabled on the cloud, you'll get to use anonymous auth locally. I think thats a reasonable thing to enable by default for this template app

sudssm: regarding --profile -- I'm going to try to release an update to edgectl that wont require the --profile on init, and just automatically creates a new one when you try to init

other/edge-server/client/README.md
```diff
@@ -9,15 +9,16 @@ React client -> Express server (Node.js Driver) -> Edge Server -> Atlas
...
+3. Download, configure, and start the Edge Server.
+4. Install Express server and React client dependencies
+5. Start the Express server and React client.
 
 This is a MERN stack that connects to the Edge Server instead of Atlas.
 In this version of the example app, MongoDB wire protocol connections connect
-directly to Edge Server without authorization.
+directly to Edge Server. You can use email/password authentication with an
```
sudssm, line 20: per my previous comment, i think the app config for this should just enable anonymous auth

other/edge-server/client/README.md
```diff
@@ -34,32 +35,34 @@ previewing Edge Server, refer to the product page at
...
 ## Quick Start
 
-1. Create an App Services App based on the `edge-server.todo` template app, and
-   get Edge Server enabled for your new app.
+1. Create an App Services App based on the `edge-server.todo` template app.
 2. Download and configure the Edge Server.
-3. Start Docker
-4. From the project's root directory, run `npm run install-deps`.
-5. Add a `.env` file inside the `node-server` directory with the details
+3. As part of the Edge Server installation, choose to create a new user. Remember
```
sudssm, line 44: you could skip this step if you choose to enable anon auth

other/edge-server/client/README.md
krollins-mdb: @sudssm, can you help me understand the use case for `--insecure-disable-auth=true`? Is it only used if people want to completely avoid a dependency on App Services?

other/edge-server/client/README.md
krollins-mdb: Nice! That's good DX. :)

other/edge-server/client/README.md
```diff
@@ -374,8 +259,23 @@ that the Edge Server is running. If you have changed the URI and/or port where
 the Edge Server is listening for wireprotocol connections, change the URI
 and/or port in your `.env` file.
 
+## Run Multiple Edge Server Instances on a Host
```
dacharyc, line 262: It looks like this section has gotten added in the middle of the "Troubleshooting" section. I see that Troubleshooting starts at ln 211 up above, and there's another "Troubleshooting" header at ln 291 below. I think _all_ of the troubleshooting errors listed here no longer apply in this version of the app, so I think we can just remove all of the troubleshooting details and keep the stuff between here and ln 289 below.

(And I think since you added the listener stuff, we can remove the part of ln 288 where we say "you can reload the app" - since you don't have to do that anymore, yay!)

other/edge-server/client/README.md
```diff
@@ -9,15 +9,15 @@ React client -> Express server (Node.js Driver) -> Edge Server -> Atlas
 
 To use this example application, you must:
 
-1. Create an App Services App based on the `edge-server.todo` template app, and
-   get Edge Server enabled for your new app.
+1. Create an App Services App based on the `edge-server.todo` template app.
```
dacharyc, line 12: With these steps here, _plus_ the quick start below, and then additional details as part of the quick start, it feels like we're providing three sets of instructions. I think I would just remove these steps, move the "Project npm commands" section down, and go directly into the quick start after the high-level summary stuff that is currently ln 17- ln 30.

other/edge-server/client/react-client/src/components/Content.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const { logoutAndDisconnect } = useAuth();
+  const { user } = useAuthContext();
+
+  const [showError, setShowError] = useState(false);
+  const [showSuccess, setShowSuccess] = useState(false);
+  const [connectionMessage, setConnectionMessage] = useState("");
+  const [connectionError, setConnectionError] = useState("");
+  const [connectionString, setConnectionString] = useState("");
+
+  // TODO: abstract all of this one level up - to <Content>.
```
dacharyc, line 34: Is this TODO still applicable? I'm inclined to say we should not merge something to the public repo that contains a "TODO" but I don't know if you've done this work and need to remove the comment, or if this work is still outstanding.

If it still needs to be done, maybe capture the info in a Jira ticket for future work and remove it from here?

dacharyc (commented):
I really love the changes you've made here - I think the automatically reloading is a big improvement, and I like the clear communication of how it is connected to the Edge Server. The ability to log in and out with different users or auth providers is really nice, too. I've got a couple of small suggestions here, and then I think we can merge this and follow up later with more improvements.

other/edge-server/client/README.md
sudssm: yeah, if you dont want to enable anonymous auth in the UI, then you can use this. its meant to speed up the time to connecting to the edge server via mongosh

though keep in mind there is a bug with anonymous auth right now

other/edge-server/backend/data_sources/mongodb-atlas/todo/Item/rules.json
```diff
@@ -3,11 +3,9 @@
...
-            "name": "readAndWriteAll",
+            "name": "edgeServerReadWriteAll",
             "apply_when": {
-                "%%user.type": {
-                    "$eq": "mtss"
-                }
+                "%%user.type": "edge"
             },
             "document_filters": {
                 "read": true,
```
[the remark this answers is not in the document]
krollins-mdb: Good idea!

other/edge-server/client/README.md
```diff
@@ -34,32 +34,35 @@ previewing Edge Server, refer to the product page at
...
   React client. You must install the Edge Server dependencies separately.
 - `start`: starts the Edge Server, Express server, and React client.
-- `shut-down`: shuts down the Edge Server, Express server, and React client.
+- `stop`: shuts down the Edge Server, Express server, and React client.
 
 ## Quick Start
 
-1. Create an App Services App based on the `edge-server.todo` template app, and
-   get Edge Server enabled for your new app.
+1. Create an App Services App based on the `edge-server.todo` template app.
```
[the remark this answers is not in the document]
krollins-mdb: Yep. Good catch! This definitely assumes a certain level of knowledge because we don't specify how to do it.

other/edge-server/client/react-client/src/components/Content.tsx
krollins-mdb: Oops, nope! I forgot to remove it. Thank you for catching!

other/edge-server/client/README.md
krollins-mdb: Yeah, as mentioned in Slack, we'll stick with what we've got until the anonymous auth bug is resolved. Thank you!

other/edge-server/client/README.md
krollins-mdb: Agreed. It was definitely confusing before. I've removed that first set of instructions.

other/edge-server/backend/data_sources/mongodb-atlas/todo/Item/rules.json
```diff
@@ -3,15 +3,15 @@
     "database": "todo",
     "roles": [
         {
-            "name": "readAndWriteAll",
+            "name": "edgeServerReadWriteAll",
```
dacharyc, line 6: If we're going to update this so the Edge Server can only write to its "own" documents, we should probably rename the rule.
```suggestion
            "name": "edgeServerReadAllWriteOwn",
```

other/edge-server/client/README.md
```diff
@@ -1,65 +1,61 @@
...
 
-1. Create an App Services App based on the `edge-server.todo` template app, and
-   get Edge Server enabled for your new app.
-2. Download and configure the Edge Server.
-3. Start Docker
-4. From the project's root directory, run `npm run install-deps`.
-5. Add a `.env` file inside the `node-server` directory with the details
+1. Create an App Services App based on the `edge-server.todo` template app. Refer
+   to [the documentation](https://www.mongodb.com/docs/atlas/app-services/reference/template-apps/#create-a-template-app) for details.
+2. [Create a Service for Edge Server](https://www.mongodb.com/docs/atlas/app-services/edge-server/manage-edge-servers/#create-a-service-for-edge-server)
```
dacharyc, line 26: If they create a template app, they don't have to also create a service - this creates a Device Sync service for them. They can just go right to creating an Edge Server instance. So I think we can remove this step and update the step numbering.

other/edge-server/client/README.md
```diff
@@ -1,65 +1,61 @@
...
+3. [Create an Edge Server Instance](https://www.mongodb.com/docs/atlas/app-services/edge-server/manage-edge-servers/#create-an-edge-server-instance)
+4. [Install and Configure the Edge Server Instance](https://www.mongodb.com/docs/atlas/app-services/edge-server/configure/#install-and-configure-the-edge-server-instance)
+5. As part of the Edge Server installation, choose to create a new user. Remember
+   the user's email and password for later.
+   Alternatively, you can skip user creation and disable authentication by running `edgectl config --insecure-disable-auth=true`.
+6. Start Docker
```
dacharyc, line 32: If you want to keep "Start Docker" in here, I'd move it before the "Install and Configure the Edge Server Instance" step. When I was just running through this with Lindsey, we found the `edgectl init` step expects Docker to be running already.

other/edge-server/client/README.md
```diff
@@ -156,147 +152,38 @@ not contain information about running the Edge Server template. You can view
...
-
-For computers running macOS, we recommend using `brew` to install required
-dependencies.
-
-### Run the Edge Server
+To set up an Edge Server:
 
-Make sure Docker is running on your local machine. Then, run the following
-command to start the Edge Server:
+1. [Create a Service for Edge Server](https://www.mongodb.com/docs/atlas/app-services/edge-server/manage-edge-servers/#create-a-service-for-edge-server)
```
dacharyc, line 162: Again, we can skip the "Create a Service" step and change the numbering on the subsequent steps if the user has created a template app.

dacharyc (approved):
Overall, LGTM! Just a couple more small tweaks for the README, and a suggestion to rename the Edge Server rule if we're going to match the restricted write permissions.

other/edge-server/backend/data_sources/mongodb-atlas/todo/Item/rules.json
krollins-mdb: Yeah, that's sensible. 😆

other/edge-server/client/README.md
krollins-mdb: D'oh, that's right. Thank you!

other/edge-server/client/README.md
krollins-mdb: Ah, right. I do want to selfishly keep it. Multiple times I've forgotten to start Docker. I know the CLI will tell people they need to do it, but I think it helps to have it in the steps.

PR #210, opened 2024-07-04
Remove Flipper

takameyer:
* In this version of React Native, current versions of Xcode are not compatible with Flipper dependencies
* This commit disable Flipper from being installed in the Pod Install phase
* In effect, this will avoid failures when using the template


takameyer:
@krollins-mdb @dacharyc 
Please review.  We have had at least two internal users who hit this issue and this was the solution.
Thanks 🙂

krollins-mdb (approved):
@takameyer, thank you for the update! LGTM.

PR #211, opened 2024-07-17
Add GH action to verify all files in /client are also in /generated

MongoCaleb:
Adds a github action to verify all files in /client are also in /generated. This ensures that the PR submitter has run bluehawk.

### DIFF
--- a/.github/workflows/verify-bluehawk.yml
+++ b/.github/workflows/verify-bluehawk.yml
@@ -1,0 +1,21 @@
+name: Verify Bluehawk Was Run
+
+on: pull_request
+
+jobs:
+  check-bluehawk:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Ensure Bluehawk Script Was Run
+        working-directory: sync-todo/v2
+        run: |
+          dir1=client/
+          dir2=generated/
+          if [ `diff $dir1 $dir2 | grep -c 'Only'` == 0 ]
+          then 
+            echo "Passed"
+          else 
+            echo "Fail -- one or more files exist in /client that aren't in /generated."
+            exit 1
+          fi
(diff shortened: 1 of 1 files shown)


.github/workflows/verify-bluehawk.yml
```diff
@@ -0,0 +1,21 @@
...
+      - name: Ensure Bluehawk Script Was Run
+        working-directory: sync-todo/v2
+        run: |
+          dir1=client/
+          dir2=generated/
+          if [ `diff $dir1 $dir2 | grep -c 'Only'` == 0 ]
+          then 
+            echo "Passed"
+          else 
+            echo "Fail -- one or more files exist in /client that aren't in /generated."
```
dacharyc, line 19: Could we provide helpful info about how to fix this failure? Maybe something like:

```suggestion
            echo "Oops - looks like you updated files in the /client directory but they don't have matching updated files in the /generated directory. Please run the `bluehawk.sh` script in `sync-todo/v2` to generate updated output files."
```

(As a bonus, it might be nice to mention _which_ files got updated in `client` but aren't present in `generated`, but if that's annoying to do, it's not essential. And I'm not sure what this suggestion looks like in output - I'm not sure if we should split the echo across multiple lines?)

.github/workflows/verify-bluehawk.yml
```diff
@@ -0,0 +1,21 @@
+name: Verify Bluehawk Was Run
```
dacharyc, line 1: Minor suggestion: running Bluehawk is an implementation detail - the thing we really want to verify is that the generated output files are present, so I think reflecting that in the check name would be more meaningful to show in the PR checks.
```suggestion
name: Verify Generated Files are Updated
```

dacharyc (approved):
This looks good! I have a couple of minor suggestions to make the check and its possible failure more meaningful for contributors.

Thanks for putting this together!

